PR #5, opened 2020-11-26
Initial migration from raven repo

cjauvin:


huard (changes requested):
Darn, I forgot that we wanted to keep large test data files outside of the repo...

Ideas ?

cjauvin:
> Darn, I forgot that we wanted to keep large test data files outside of the repo...
> Ideas ?

We could include them as an optional extra package maybe?

Zeitsperre:
@cjauvin There's already some discussion about this. I'll likely be creating a tesdata repo for Raven(Py) later today.

Zeitsperre:
For more information: https://github.com/Ouranosinc/raven/issues/318

tlvu:
> > Darn, I forgot that we wanted to keep large test data files outside of the repo...
> > Ideas ?
> 
> We could include them as an optional extra package maybe?

FYI, the commit that has those large testdata should be destroyed and recreated.  Just commit again to delete those large testdata file will still keep them in the history and bloat the repo size.

cjauvin:
yes absolutely, good idea, will do!

cjauvin:
Actually I think that this commit crashed my text editor (Emacs).. 😟

cjauvin:
I removed the `tests/testdata` folder from my previous commit and force-pushed the change.

huard (approved):

Issue #4, opened 2020-11-26
Close netCDF files when overwriting existing files with the same name


huard:
### Description

With xarray 0.16.1, overwriting a file on disk with the same name does not clear the cache. So when opening a file with the same name, it fetches the cached file rather than the new file on disk. 

Possible solutions: 
- Given unique name to model outputs (.e.g uuid)
- Call `.close()` on all opened netCDF files when `overwrite=True`.

Zeitsperre:
@huard Could this be part of the issue we have with test instability?

huard:
It may. You could check by running tests sequentially (no xdist)

PR #6, opened 2020-11-27
Fix problem with netCDF file caching

cjauvin:
This is a possible fix for https://github.com/CSHS-CWRA/RavenPy/issues/4. In order to use the `with` statement, I'm using a generator, which I renamed to `hydrographs`, and introduced an additional `hydrograph` property, in order to keep the tests `test_emulators.py::TestGR4JCN` from crashing. As is, this fixes the problem we had with `test_overwrite` and `xarray==0.16.1`.

Note that this PR is just intended as a discussion, as I'm really not sure that this is the solution we want/need in this context. Also, whatever the solution we adopt, it will need to be expanded to the other uses of netCDF files throughout the project, to be consistent.


huard (commented):
Elegant, but I'm wondering when exactly is the file closed ?

cjauvin:
> Elegant, but I'm wondering when exactly is the file closed ?

The `__exit__` method of the context manager is called at every iteration of the for loop, and thus every time `next()` is called on the generator. So the result I believe is a collection (or singleton) of properly closed file handles.

huard:
But when you `get` the property, don't you create an other generator object, so the `next` is called not on the first object, but on the second object ?

cjauvin:
My understanding that accessing the property is simply refering to the generator itself (i.e. the property/method _is_ the generator) and that while you haven't "done" anything with it, no file is either open or close (because no `yield` has been executed yet). But I will verify to make sure.

cjauvin:
@david:

``` python
class Ressource:
    def __init__(self, i):
        self.i = i
        print(f"init {self.i}")

    def __enter__(self):
        print(f"enter {self.i}")
        return self.i

    def __exit__(self, *exc):
        print(f"exit {self.i}")


class User:
    @property
    def ressources(self):
        for i in range(3):
            with Ressource(i) as r:
                yield r


u = User()
rs = [r for r in u.ressources]

# init 0
# enter 0
# exit 0
# init 1
# enter 1
# exit 1
# init 2
# enter 2
# exit 2
```

huard:
But then I assume that `close` is called, so how can we access the content then ? There may be some xarray cache magic here.

cjauvin:
I have no idea, and it might be why this solution is not really appropriate actually.

huard:
I suggest we merge this and see how it goes.

huard:
Should we merge this ? Seems we're having problems with caching.

cjauvin:
This is for `hydrographs` only, do you think this is the only one causing issue? If not this trick should be repeated elsewhere to be consistent.

huard:
Same trick should be applied to storage as well.

cjauvin:
I think it's better to close this as too much time has passed and I'm not sure the issue is up to date anymore.

PR #13, opened 2020-12-06
RVH Support

cjauvin:
This is a first pass for the addition of RVH support in RavenPy.

A few things to note:
1. I have rewritten the logic to extract the data from the HRU Routing Product shapefile (as it is right now in the `BasinMaker`, aka. Routing Toolbox project) and encapsulated it in this specialized "importer" class: https://github.com/CSHS-CWRA/RavenPy/blob/feature/rvh-support/ravenpy/models/importers.py#L6
2. In its current state, I estimate the import logic to be about ~90% correct; it's not perfect, but it's rather close
3. Apart from the importer, which is a new abstraction (and thus very open to criticism or change), I have tried to stay in the spirit of David's existing code and design ideas as much as possible

As we discussed, a possible strategy to make sure that this gets nicely integrated in the `BasinMaker` project (and stay DRY as much as possible) would be to show this work, and offer to use it as a replacement for the corresponding logic: concretely, the way I see it, `BasinMaker` would import `ravenpy`, and instantiate a `ravenpy.models.rv.RVH` class, and it would also make use of my `ravenpy.models.importers.RoutingProductShapefileImporter`.


ravenpy/models/state.py
```diff
@@ -142,3 +144,58 @@ class BasinStateVariables(NamedTuple):
...
+    """Record to populate RVH :SubBasins table."""
+
+    subbasin_id: int = 0
+    name: str = "subXXX"
+    downstream_id: int = 0
+    profile: str = "chn_XXX"
+    reach_length: float = 0
+    gauged: bool = False
+
+    def render_for_raven(self):
```
huard, line 159: Rename to `to_rv` ?

ravenpy/models/global/global.rvh
```diff
@@ -0,0 +1,21 @@
...
+
+:SubBasinGroup    land_subbasin_group
+  {txt_land_subbasin_group_ids}
+:EndSubBasinGroup
+
+:SubBasinGroup    lake_subbasin_group
+  {txt_lake_subbasin_group_ids}
+:EndSubBasinGroup
+
+{txt_lakes}
```
huard, line 21: The HBV-EC model has another directive called `:SubBasinProperties`. If the plan is to replace the individual emulators file template, I suspect we'll have to add that here. I suggest we first merge this however and then work on emulator integration.

ravenpy/models/importers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    MAX_RIVER_SLOPE = 0.00001
+    USE_LAKE_AS_GAUGE = False
+    WEIR_COEFFICIENT = 0.6
+    USE_LAND_AS_GAUGE = False
+
+    def __init__(self, path):
+        self._df = geopandas.read_file(path)
+
+    def extract(self):
```
huard, line 23: Add docstring.

ravenpy/models/rv.py
```diff
@@ -712,6 +712,90 @@ def txt_basin_state(self):
         return "\n".join(txt).replace("[", "").replace("]", "")
 
 
+class RVH(RV):
+    def __init__(self, importer, **kwargs):
```
huard, line 716: This is a *smart* RVH class, in the sense that the data import logic is embedded into it. This is different from the other RV subclasses, whose only logic is how to convert basic python objects to RV content (the smart logic is upstream, mostly in the `base.Raven` class. Some of the *smart* logic at the moment lives inside the `Raven` class, but I suspect that we'll need to rethink this. At the moment, the `Raven` class accepts `latitude`, `longitude`, `area`, scalar arguments, but this might later clash with the distributed configuration. 

I think we should merge this as is for now. 

Once RavenPy becomes a dependency of raven, then maybe we can start thinking about a refactoring. Maybe think about a `Watershed` class, with a `Distributed` subclass (which could be instantiated with the `importer` instance), and a `Global` subclass (with a simple instantiation similar to what is done right now within the `Raven` class). `Raven` would require as an input an instance of `Watershed`.

huard (approved):

ravenpy/models/state.py
cjauvin: done.

ravenpy/models/importers.py
cjauvin: done

PR #14, opened 2020-12-07
Add support for Channel Profile (RVP)

cjauvin:
This adds support for the import and generation of the `channel_properties.rvp` (aka. Channel Profiles).

@dustming In particular take a look if you can at: 

https://github.com/CSHS-CWRA/RavenPy/blob/feature/channel-profile-rvp-support/ravenpy/models/importers.py#L127

which is a reimplementation of your: 

https://github.com/dustming/basinmaker/blob/master/BasinMaker/processing_functions_raven_model_io.py#L585


ravenpy/models/rv.py
```diff
@@ -796,6 +789,30 @@ def txt_hrus(self):
         return "\n".join(txt)
 
 
+class RVP(RV):
+    def __init__(self, channel_profiles, **kwargs):
+        self._channel_profiles = channel_profiles
+        super().__init__(**kwargs)
+
+    @property
+    def channel_profiles(self):
```
huard, line 798: Note that in the rvp files for the emulators, there is a lot of stuff beyond the channel profile. Here, `RVP.render_to_rv` only renders a small part of the RVP. What I mean is that this is the renderer for the channel profile only, not the full RVP. 

Elsewhere in the rv code, I've sometimes used properties to actually print to template, that is, the `render_to_rv` is built in the property `getter`. For example , `RVI.supress_output` directly outputs the string to be inserted into the template. I don't think that was such a good idea, because the RV attributes are now split in two: those that return a Python object that is converted to a string by the templating, and those that return directly a string. I think this was a bad design decision done for a quick fix. Another design pattern I've used is in RVC with the `txt_basin_state`. There, you have both the python objects (hru_state, basin_state), and the `txt_hru_sate`, `txt_basin_state` doing the rendering. I think this could work here: 
- `channel_profiles` is a list or dictionary of channel profile objects
- `txt_channel_profiles` does the `to_rv` rendering. 

I'm sure we could improve on that in the future.

ravenpy/models/rv.py
```diff
@@ -881,7 +898,15 @@ def _parser(lines, indent="", fmt=str):
...
-                out[old_key] = _parser(itertools.chain([line,], lines), new_indent)
+                out[old_key] = _parser(
+                    itertools.chain(
+                        [
+                            line,
+                        ],
+                        lines,
+                    ),
+                    new_indent,
+                )
```
huard, line 909: black seems to be aggressive on this one.

huard (changes requested):

ravenpy/models/rv.py
cjauvin: I fixed it. Note that the reason for this extremism was the pre-existing trailing comma for this list: `[line,]`

> A pre-existing trailing comma informs Black to always explode contents of the current bracket pair into one item per line. Read more about this in the Pragmatism section below.

ravenpy/models/rv.py
cjauvin: Yes it makes sense, and that is exactly the pattern that I have used for the `RVH` class in my previous PR.

ravenpy/models/rv.py
```diff
@@ -712,84 +712,80 @@ def txt_basin_state(self):
 
 
 class RVH(RV):
```
huard, line 714: If we don't need special setters and getters, these can be dumb class attributes, no?

cjauvin: Some getters implement special twists though:

https://github.com/CSHS-CWRA/RavenPy/blob/feature/channel-profile-rvp-support/ravenpy/models/rv.py#L732-L742

huard: Right, looked too quickly.

huard (approved):

huard (approved):
This can be done in a future PR, but RVH will also need to implement `SubBasinProperties`. See `raven-hbv-ec.rvh`.

Issue #17, opened 2020-12-10
Treat Anaconda environments as virtualenvs


Zeitsperre:
The current behaviour when performing an installation (`python setup.py install`) in an Anaconda environment is to raise an error. Anaconda should be recognized as an acceptable implementation of python as it mimics virtualenv if managed properly.

cjauvin:
I made a 180 deg turn and I'm now bullish on :snake: !

PR #19, opened 2020-12-15
Add grid weight import mechanism (for RVT)

cjauvin:
This implements https://github.com/julemai/GridWeightsGenerator as a specialized class in `ravenpy.models.importers`, and connects it to the `ravenpy.models.rv.RVT` class.

Note that I have made many changes to the algorithm itself, but I have kept the original comments in their context when it's applicable.


setup.py
```diff
@@ -80,14 +81,14 @@ def install_binary_dep(self, url, name, rev_name, binary_name, make_target=""):
         )
 
     def run(self):
-        if sys.base_prefix == sys.prefix:
+        if sys.base_prefix == sys.prefix and not os.getenv("CONDA_PREFIX"):
```
Zeitsperre, line 84: Is this enough to close #17 ?

cjauvin: I think it might be, but this is a bit experimental and I'm not 100% sure it covers all edge cases.

ravenpy/models/importers.py
```diff
@@ -197,3 +204,285 @@ def _extract_hru(self, row) -> HRUsCommandRecord:
...
+
+    """
+    The original version of this algorithm can be found at: https://github.com/julemai/GridWeightsGenerator
+    """
+
+    CRS_LLDEG = 4326  # EPSG id of lat/lon (deg) coordinate referenence system (CRS)
+    CRS_CAEA = 3573  # EPSG id of equal-area    coordinate referenence system (CRS)
+    HRU_FIELD = "HRU_ID"
+    DIM_NAMES = ("lon", "lat")
+    VAR_NAMES = ("lon", "lat")
```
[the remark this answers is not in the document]
cjauvin: Yes they might well be, but for now I have treated all those values as "I'm not sure what needs to be eventually parametrized or not, how and by whom", so I have gathered them as importer-level class variables as a temporary measure.

cjauvin:
@huard This https://github.com/CSHS-CWRA/RavenPy/commit/e7fdcfb7435cf7428389a1d9c6c1c6ecaf452576 implements a demoable starting point for what we discussed this afternoon (and will need to discuss tomorrow).

cjauvin:
I'm merging this as I want to press forward with 
* testdata
* raven / ravenpy modularization
* upload to pypi (to test it works, reserve the RavenPy name and hopefully provide a user-friendly to wrap @julemai 's grid weight script (which is part of this PR))

PR #21, opened 2020-12-16
Add environment.yml and xskillscore and climpred to dependencies

Zeitsperre:
This addresses a desired library identified in https://github.com/Ouranosinc/raven/issues/336. I'm merging it based off of this branch, as the behaviour/enhancement (close-netcdf-files) is needed for it.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,14 @@
+name: RavenPy
+channels:
+- conda-forge
+- defaults
+dependencies:
+- Click>=7.0
+- gdal==3.0.4
+- rasterio
+- statsmodels
+- xarray==0.16
+- xclim==0.18
+- wheel
+- xskillscore
+- climpred>=2.1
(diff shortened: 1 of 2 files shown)


cjauvin:
I'm a little confused and didn't follow super well. I'm not sure however that the solution I devised in https://github.com/CSHS-CWRA/RavenPy/pull/6/files is the way to go. I don't know the answer to the question raised (when exactly is the file closed, and what consequence does it have?)

huard:
I made the call. Empirically, it works, and we have tests to ensure validity, so I'm willing to run with it.

cjauvin:
Ok but I'm just confused about the relationship between this "bug" (if it's really one) and the exact version of xarray that we're talking about.. could someone clarify?

huard:
climpred requires xarray 0.16.2. The cache issue affects xarray 0.16.1 and 0.16.2. Your PR solving the cache bug allows us to unpin xarray 0.16.0 and install climpred.

huard (approved):

PR #22, opened 2020-12-21
Use bumpversion

Zeitsperre:
`Bumpversion` is a tool we use to get around the problem of tracking/updating versions between releases.

https://github.com/c4urself/bump2version

The API is quite simple: `bumpversion major/minor/patch` will increment the semver part that you are interested in advancing. I have ported a piece of configuration that allows one to set a version as "release-ready" as well. Bumping major will set minor and patch both to `0`, minor will reset patch to `0`. It's pretty no-nonsense.

I have disabled tagging and committing by bumpversion since it doesn't play very well with `pre-commit` and `black` sometimes, but the process of releasing/tagging is not complicated is not complicated:

In other words, to create a new tagged release:
```
$ bumpversion minor
$ git commit -m "bumped version to vX.Y+1.Z-beta"
$ bumpversion release
$ git commit -m "bumped version to vX.Y+1.Z"
$ git tag "vX.Y+1.Z"
$ git push (--tags)
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,12 +4,12 @@ commit = False
 tag = False
 
 [bumpversion:file:setup.py]
-search = version='{current_version}'
-replace = version='{new_version}'
+search = version="{current_version}"
+replace = version="{new_version}'"
 
-[bumpversion:file:ravenpy/__init__.py]
-search = __version__ = '{current_version}'
-replace = __version__ = '{new_version}'
+[bumpversion:file:ravenpy/__version__.py]
+search = __version__ = "{current_version}"
+replace = __version__ = "{new_version}"
 
 [bumpversion:file:.cruft.json]
 search = "version": "{current_version}",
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -1,39 +1,48 @@
...
 
 [aliases]
-# Define setup.py command aliases here
 test = pytest
 
 [tool:pytest]
 collect_ignore = ['setup.py']
 addopts = --verbose
-filterwarnings =
+filterwarnings =
```
cjauvin, line 34: Your editor seems to be adding an extra space here? Mine is configured to remove it..

cjauvin (approved):
I like it, it's an excellent idea, thanks! So it's ready to go as is?

setup.cfg
Zeitsperre: Oh, it's a strange bug with bump2version. It's harmless so long as we don't commit on bump. https://github.com/c4urself/bump2version/issues/124

PR #23, opened 2020-12-22
Port of latest grid weight generation script features

cjauvin:
This is a port of @julemai's latest grid weight generation script features into RavenPy. There are many things open to discussion!

### DIFF
--- a/ravenpy/cli.py
+++ b/ravenpy/cli.py
@@ -99,11 +99,12 @@ def generate_grid_weights(
     Generate grid weights in various formats.
 
     INPUT_FILE_PATH: Either (A) Example NetCDF file containing at least 1D or 2D latitudes and 1D or 2D longitudes where grid
-    needs to be representative of model outputs that are then required to be routed. Or (B) a shapefile that contains shapes
-    of subbasins and one attribute that is defining its index in the NetCDF model output file (numbering needs to be [0 ... N-1]).
+    needs to be representative of model outputs that are then required to be routed. Or (B) a shapefile (either a .shp or a .zip)
+    that contains shapes of subbasins and one attribute that is defining its index in the NetCDF model output file (numbering
+    needs to be [0 ... N-1]).
 
-    ROUTING_FILE_PATH: Shapefile that contains all information of the routing toolbox for the catchment of interest (and maybe some
-    more catchments).
+    ROUTING_FILE_PATH: Shapefile (either a .shp or a .zip) that contains all information of the routing toolbox for the catchment
+    of interest (and maybe some more catchments).
 
     The script outputs the results (in the chosen format) on STDOUT, so they can be redirected to a file using the `>` shell operator.
     """
--- a/ravenpy/cli.py
+++ b/ravenpy/cli.py
@@ -25,7 +25,7 @@ def cli():
     type=click.Tuple([str, str]),
     default=RoutingProductGridWeightImporter.DIM_NAMES,
     show_default=True,
-    help="Ordered dimension names of longitude (x) and latitude (y) in the NetCDF file.",
+    help="Ordered dimension names of longitude (x) and latitude (y) in the NetCDF file. Example: 'rlon rlat'.",
 )
 @click.option(
     "-v",
@@ -34,7 +34,7 @@ def cli():
     type=click.Tuple([str, str]),
     default=RoutingProductGridWeightImporter.VAR_NAMES,
     show_default=True,
-    help="(A) Variable name of 2D longitude and latitude variables in NetCDF (in this order). Example: 'lon,lat'. Or (B) Attribute name in input_file shapefile (option -i) that defines the index of the shape in NetCDF model output file (numbering needs to be [0 ... N-1]).",
+    help="Variable name of 1D or 2D longitude and latitude variables in NetCDF (in this order). Example: 'lon lat'.",
 )
 @click.option(
     "-c",  # legacy script short option
@@ -42,7 +42,7 @@ def cli():
     type=str,
     default=RoutingProductGridWeightImporter.ROUTING_ID_FIELD,
     show_default=True,
-    help="Name of column in routing information shapefile containing unique key for each dataset.",
+    help="Name of column in routing information shapefile containing unique key for each dataset. Example: 'HRU_ID'.",
 )
 @click.option(
     "-f",  # legacy script short option
@@ -50,7 +50,7 @@ def cli():
     type=str,
     default=RoutingProductGridWeightImporter.NETCDF_INPUT_FIELD,
     show_default=True,
-    help="Attribute name in input_file shapefile that defines the index of the shape in NetCDF model output file (numbering needs to be [0 ... N-1]).",
+    help="Attribute name in INPUT_FILE_PATH shapefile that defines the index of the shape in NetCDF model output file (numbering needs to be [0 ... N-1]). Example: 'NetCDF_col'",
 )
 @click.option(
     "-g",
@@ -58,7 +58,7 @@ def cli():
     multiple=True,
     type=str,
     show_default=True,
-    help="Basins of interest (corresponds to 'Gauge-ID' in the shapefile).",
+    help="Streamflow gauge ID of interest (corresponds to 'Gauge_ID' in the shapefile). Can be a comma-separated list of multiple. Example: '02LB005,02LB008'",
 )
 @click.option(
     "-s",
@@ -66,7 +66,7 @@ def cli():
     multiple=True,
     type=str,  # to support alternate "-s 123,456" legacy syntax
     show_default=True,
-    help="Sub IDs of most downstream subbasins (containing usually a gauge station, corresponds to 'SubId' in the shapefile).",
+    help="IDs of subbasins of interest (containing usually a gauge station, corresponds to 'SubId' in the shapefile). Can be a comma-separated list of multiple. Example: '7399,7400'",
 )
 @click.option(
     "-e",  # legacy script short option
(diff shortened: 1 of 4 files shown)


cjauvin:
New features:
* Support for 1d coords
* Support for NC input file

ravenpy/cli.py
```diff
@@ -34,13 +34,23 @@ def cli():
     type=click.Tuple([str, str]),
     default=RoutingProductGridWeightImporter.VAR_NAMES,
     show_default=True,
-    help="Ordered variable names of longitude (x) and latitude (y) in the NetCDF file.",
+    help="(A) Variable name of 2D longitude and latitude variables in NetCDF (in this order). Example: 'lon,lat'. Or (B) Attribute name in input_file shapefile (option -i) that defines the index of the shape in NetCDF model output file (numbering needs to be [0 ... N-1]).",
```
julemai, line 37: Revert to previous version as we now have the `routing-id-field` as separate option. Please leave "longitude (x)"  and "latitude (y)". It's easier for some people to refer to x/y.

cjauvin:
Still to do (in the context of this PR):
* Implement the error reporting at the end of the generation script
* Add tests for `-s` and `-g` options

ravenpy/cli.py
```diff
@@ -34,13 +34,23 @@ def cli():
...
     show_default=True,
+    help="Name of column in routing information shapefile containing unique key for each dataset.",
+)
+@click.option(
+    "-f",  # legacy script short option
+    "--netcdf-input-field",
+    type=str,
+    default=RoutingProductGridWeightImporter.NETCDF_INPUT_FIELD,
+    show_default=True,
+    help="Attribute name in input_file shapefile that defines the index of the shape in NetCDF model output file (numbering needs to be [0 ... N-1]).",
```
julemai, line 53: Maybe: `input_file` --> `INPUT_FILE_PATH `

ravenpy/models/importers.py
```diff
@@ -219,78 +219,82 @@ class RoutingProductGridWeightImporter:
...
+        elif input_file_path.suffix == ".shp":
+            self._input_data = geopandas.read_file(input_file_path)
+            self._input_is_netcdf = False
+        else:
+            raise ValueError(
+                "The input file must be a shapefile (.shp or .zip) or NetCDF"
+            )
+
+        routing_file_path = Path(routing_file_path)
+        if routing_file_path.suffix == ".zip":
```
julemai, line 271: Should there be an `elif` for ".shp" and a `ValueError` if it's neither?

ravenpy/models/importers.py
```diff
@@ -420,18 +379,191 @@ def extract(self) -> GridWeightsCommand:
...
+            if len(lon_dims) == 2 and len(lat_dims) == 2:
+                if lon_dims == self._dim_names:
+                    lon_var = np.transpose(lon_var)
+
+                if lat_dims == self._dim_names:
+                    lat_var = np.transpose(lat_var)
+
+            elif len(lon_dims) == 1 and len(lat_dims) == 1:
+                # Coord vars are 1d, make them 2d
+                lon_var_size = lon_var.size
```
julemai, line 445: needed?

ravenpy/models/importers.py
```diff
@@ -420,18 +379,191 @@ def extract(self) -> GridWeightsCommand:
...
+            for ishape in range(self._nlat):
+
+                idx = np.where(self._input_data[self._netcdf_input_field] == ishape)[0]
+                if len(idx) == 0:
+                    # print(
+                    #     "Polygon ID = {} not found in '{}'. Numbering of shapefile attribute '{}' needs to be [0 ... {}-1].".format(
+                    #         ishape, input_file, key_colname_model, nshapes
+                    #     )
+                    # )
+                    raise ValueError("Polygon ID not found.")
```
julemai, line 549: Should we be a bit more talkative here telling the user which ID can't be found where and in which attribute column?

ravenpy/models/importers.py
```diff
@@ -420,18 +379,191 @@ def extract(self) -> GridWeightsCommand:
...
+                if len(idx) > 1:
+                    # print(
+                    #     "Polygon ID = {} found multiple times in '{}' but needs to be unique. Numbering of shapefile attribute '{}' needs to be [0 ... {}-1].".format(
+                    #         ishape, input_file, key_colname_model, nshapes
+                    #     )
+                    # )
+                    raise ValueError("Polygon ID not unique.")
```
julemai, line 556: More talkative?

tests/test_cli.py
```diff
@@ -5,20 +5,55 @@
...
 
-    result = runner.invoke(generate_grid_weights, ds)
+    result = runner.invoke(generate_grid_weights, params)
 
     assert result.exit_code == 0
     assert not result.exception
     assert ":NumberHRUs 51" in result.output
     assert ":NumberGridCells 100" in result.output
     assert len(result.output.split("\n")) == 216
+
```
julemai, line 23: I think we need to test at least one weight that is derived..

tests/test_cli.py
```diff
@@ -5,20 +5,55 @@
...
+    ]
+    params = map(str, params)
+
+    result = runner.invoke(generate_grid_weights, params)
+
+    assert result.exit_code == 0
+    assert not result.exception
+    assert ":NumberHRUs 51" in result.output
+    assert ":NumberGridCells 9801" in result.output
+    assert len(result.output.split("\n")) == 128
```
julemai, line 42: I think we need to test at least one weight that is derived..

tests/test_cli.py
```diff
@@ -5,20 +5,55 @@
...
+        get_test_data("raven-routing-sample", "finalcat_hru_info.zip")[0],
+    ]
+    params = map(str, params)
+
+    result = runner.invoke(generate_grid_weights, params)
+
+    assert result.exit_code == 0
+    assert not result.exception
+    assert ":NumberHRUs 51" in result.output
+    assert ":NumberGridCells 810" in result.output
```
julemai, line 58: I think we need to test at least one weight that is derived..

ravenpy/models/importers.py
cjauvin: We could but since the CLI specifies that it must be a shapefile, if it's not we can just let the call to `geopandas.read_file` crash (this is a bit lazy of course).

Zeitsperre: There isn't a special mimetype-like prepending needed for a shape. `geopandas.read_file()` should be able to handle them. Raising a `ValueError` will cause errors that otherwise wouldn't be raised (e.g. user sends a valid GeoJSON, GeoPackage, SQLite, etc.)

tests/test_cli.py
```diff
@@ -5,20 +5,55 @@
...
+    result = runner.invoke(generate_grid_weights, params)
+
+    assert result.exit_code == 0
+    assert not result.exception
+    assert ":NumberHRUs 51" in result.output
+    assert ":NumberGridCells 9801" in result.output
+    assert len(result.output.split("\n")) == 128
+
+
+def test_generate_grid_weights_with_shp_input():
```
julemai, line 45: This test runs with the default `AREA_ERROR_THRESHOLD=0.05` which will result in several subbasins where weights do not sum up to 1.

I think it would be good to have a test with having `AREA_ERROR_THRESHOLD=0.5` and checking that one of those basins is adjusted correctly.

ravenpy/models/importers.py
cjauvin: Yes because I reassign `lon_var` to itself on the next line, which would cause the line after the next (`lat_var = ..`) to be buggy, without this memorized value.

ravenpy/models/importers.py
cjauvin: Yes absolutely, this is part of the additional error reporting that I want to implement in this PR. But I wanted to open it early, for you to see.

tests/test_cli.py
cjauvin: If you want I can help you setting up to run (and write) tests.

ravenpy/models/importers.py
```diff
@@ -219,78 +219,82 @@ class RoutingProductGridWeightImporter:
...
         # WGS 84 / North Pole LAEA Canada
-        self._shapes = self._shapes.to_crs(
+        self._routing_data = self._routing_data.to_crs(
             epsg=RoutingProductGridWeightImporter.CRS_CAEA
         )
 
-        self._shapes = self._shapes.drop_duplicates(self._hru_id_field).sort_values(
-            self._hru_id_field
-        )
+        self._routing_data = self._routing_data.drop_duplicates(
```
julemai, line 286: I am unsure about dropping duplicates here...

tests/test_cli.py
```diff
@@ -5,20 +5,55 @@
...
-    result = runner.invoke(generate_grid_weights, ds)
+    result = runner.invoke(generate_grid_weights, params)
 
     assert result.exit_code == 0
     assert not result.exception
     assert ":NumberHRUs 51" in result.output
     assert ":NumberGridCells 100" in result.output
     assert len(result.output.split("\n")) == 216
+
+
```
julemai, line 24: I think we need an additional test where we invoke multiple gauges to check that removing duplicates etc is working properly.

ravenpy/models/importers.py
julemai: I am suggesting an additional test below...

ravenpy/models/importers.py
julemai: ok. Was just wondering because we also distinguish between zip and shp for the `input_file_path`

ravenpy/models/importers.py
julemai: Absolutely!

ravenpy/models/importers.py
cjauvin: @Zeitsperre I'm pretty certaint that `geopandas.read_file` crashes if I do not prepend a `zip://` schema when required though.. you say it does not?

ravenpy/models/importers.py
cjauvin: You seem to be doing it in your own code though: https://github.com/julemai/GridWeightsGenerator/blob/main/derive_grid_weights.py#L453-L463

This is just the idiomatic numpy way of doing it.

ravenpy/models/importers.py
Zeitsperre: No, you are correct, it will crash. I'm referring to why it isn't necessary to perform similar formatting for other files.

ravenpy/models/importers.py
julemai: yes, but _after_ I have build the tree...

cjauvin: ok so you think we should tolerate ROUTING_SHAPEFILE rows having the same HRU_ID?

julemai: Yes, it's a long story. There is nothing to resolve this for now. I hate this too but...

tests/test_cli.py
julemai: See: "test_generate_grid_weights_with_weight_rescaling()"

tests/test_cli.py
julemai: Added

PR #29, opened 2021-01-13
Remove the venv constraint for setup

cjauvin:
After having had some trouble with the RTD build, I revisited the venv constraint for setup, and here's my proposition.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,4 +5,5 @@ sphinx:
 
 python:
   install:
-    - requirements: requirements_docs.txt
+    - method: pip
+      path: .[docs]
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,5 +5,4 @@ sphinx:
 
 python:
   install:
-    - requirements: requirements.txt
     - requirements: requirements_docs.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,8 @@
+version: 2
+
+sphinx:
+  configuration: docs/conf.py
+
+python:
+  install:
+    - requirements: requirements_docs.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,4 +5,5 @@ sphinx:
 
 python:
   install:
+    - requirements: requirements.txt
     - requirements: requirements_docs.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,5 @@ version: 2
 sphinx:
   configuration: docs/conf.py
 
-python:
-  install:
-    - method: pip
-      path: .[docs]
+conda:
+  environment: environment-docs.yml
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,4 +4,4 @@ sphinx:
   configuration: docs/conf.py
 
 conda:
-  environment: environment-docs.yml
+  environment: environment-rtd.yml
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -9,12 +9,11 @@ Stable release
 --------------
 
 Because RavenPy relies on the `Raven <http://raven.uwaterloo.ca>`_ binary (which is downloaded and
-compiled during the setup), it must be installed in a virtual
+compiled during the setup), it is preferable to install it in a virtual
 environment (either a classic `Python venv
 <https://docs.python.org/3/tutorial/venv.html>`_ or a `Conda
 environment
-<https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_
-(which must be different than the ``base`` environment).
+<https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_.
 
 As the ``GDAL`` Python library requires some system-level dependencies
 to be properly compiled, you will probably have to do:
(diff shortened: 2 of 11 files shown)


docs/installation.rst
```diff
@@ -9,12 +9,11 @@ Stable release
 --------------
 
 Because RavenPy relies on the `Raven <http://raven.uwaterloo.ca>`_ binary (which is downloaded and
-compiled during the setup), it must be installed in a virtual
+compiled during the setup), it is preferable to install it in a virtual
 environment (either a classic `Python venv
 <https://docs.python.org/3/tutorial/venv.html>`_ or a `Conda
 environment
-<https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_
-(which must be different than the ``base`` environment).
```
tlvu, line 17: Just curious, why remove this notice "which must be different than the ``base`` environment" ?

huard (approved):
All good !

tlvu (commented):
Just a curious comment.

docs/installation.rst
Zeitsperre: In general, it's not great practice to muck up the `base` environment. It's the same principle as the demand for a `venv` but applied to Anaconda.

docs/installation.rst
```diff
@@ -43,38 +42,47 @@ path, and that you can verify by doing:
...
+   $ pip install --verbose --editable .
 
-If you want to run the tests, install with:
+If you want to run the tests, install the requirements:
 
 .. code-block:: console
 
-   $ cd /path/to/ravenpy
-   $ pip install . --verbose --install-option="--with-raven" --install-option="--with-testdata"
    $ pip install -r requirements_dev
```
Zeitsperre, line 76:
```suggestion
   $ pip install --editable .[dev]
   or 
   $ pip install -r requirements_dev
```

Zeitsperre (approved):

docs/installation.rst
cjauvin: Yes good catch. The reason for this particular change is that this PR removes the notion of "focing the user of being in a venv (conda or plain)". Since there is no such verification, it means that I no longuer have to check if the conda env is the base one.

This is a complicated thought process (of mine). My primary reason for this PR was the RTD build: it failed because if wasn't in a venv. So I revisited the install process and came up with this PR. So I discovered something new:

https://github.com/CSHS-CWRA/RavenPy/blob/setup-without-venv/setup.py#L94-L97

So even though the fact we are removing the venv constraint might result in unwanted results in some specific contexts (such as the user installing with sudo and his system python) I think this PR still provides reasonable outcomes (i.e. the binaries installed in the venv/bin or ~/.local/bin) in most cases.

What do you think? This is all very experimental to me, I might be wrong for many assumptions.

docs/installation.rst
cjauvin: Didn't know I could do it like that, I like it!

Zeitsperre: If users have a non-standard shell (`zsh`, `fish`), it needs to be `".[dev]"`, but if they're using a non-standard shell, they probably know this already.

docs/installation.rst
tlvu: Still not 100% clear what your RTD build problem is but I can say that for Raven, the Makefile targets do not have dependencies on Conda env creation targets anymore.

Basically it assumes user already managed to have a working env.  I don't remember if there is detection for user not in any kind of venv which would allow modifying the system python, which is bad.

However, make targets to create the conda env is left there for user to call manually if desired (saving the user from remembering the exact conda syntax and avoid using the `base` conda env).

So basically, no direct management of pip/conda venv is OK with me.  We should leave enough help behind to assist user to create the venv though, which I think you did via documentation.  I would prefer direct in some code to minimize user error but docs is fine too.

cjauvin: The central problem we have to manage in this context is the handling of external binaries (Raven and OSTRICH), that need to be available (on the PATH or otherwise) to RavenPy.

.readthedocs.yml
```diff
@@ -5,4 +5,5 @@ sphinx:
 
 python:
   install:
-    - requirements: requirements_docs.txt
+    - method: pip
+      path: .[docs]
```
Zeitsperre, line 9:
```suggestion
      path: .
      extra_requirements:
        - docs
```

cjauvin:
ok guys @huard @Zeitsperre I think I'm now satisfied.. I took a big conda turn, for the best I hope.. what do you think?

docs/installation.rst
```diff
@@ -59,30 +67,31 @@ The sources for RavenPy can be obtained from the GitHub repo:
...
 
    $ cd /path/to/ravenpy
-   $ pip install --verbose --editable .
-
-If you want to run the tests, install the requirements:
+   $ conda env create -f environment.yml
+   $ conda activate ravenpy-env
 
-.. code-block:: console
+You can then install RavenPy with:
```
Zeitsperre, line 84:
```suggestion

.. code-block:: console

```

Zeitsperre:
Is there a way to enable the checks for RtD on PRs for this repo? It would be good to ensure that the next PR doesn't break all of @cjauvin 's hard work!

docs/installation.rst
cjauvin: :eagle: :eye: I like that thanks!

cjauvin:
@Zeitsperre actually I guess the idea would be to activate a build only with change on master? now that we know that it works it's enough I guess to only verify manually.. not sure

Zeitsperre:
We don't lose anything by having builds on PR in addition to push. It's surprisingly easy to break the docs, haha.

PR #30, opened 2021-01-13
Routing model emulator

huard:
Adds configuration file for routing-only simulations. 

This is going to be used to run the distributed modeling tutorial, and used as a stepping stone to integrate routing capability in other models. 

TODO: 
- [ ] Add tests based on distributed model tutorial
- [x] Complete templating refactoring to be able to write new-style `Commands` to the templates. 
- [ ] Write derived_parameters logic to compute `avg_annual_runoff` required in `rvp`


huard:
@cjauvin I suggest we open another PR to harmonize template formatting. Then come back to this PR.

cjauvin:
@huard @julemai The unit test reproducing the Lievre tutorial with the Routing Emulator is now fully functional.

tests/test_emulators.py
```diff
@@ -1060,3 +1072,142 @@ def test_simple(self):
...
+
+        model = Routing()  # "/tmp/ravenpy_routing_emu_dev")
+
+        #######
+        # RVI #
+        #######
+
+        streaminputs = xr.open_dataset(vic_streaminputs_nc_path)
+
+        start_date = streaminputs.time[0].dt
```
huard, line 1134: streaminputs.indexes['time'][0] might give you something easier to work with.

tests/test_emulators.py
```diff
@@ -1060,3 +1072,142 @@ def test_simple(self):
...
+            reservoirs,
+            channels,
+            hrus,
+        ) = rvh_importer.extract()
+
+        model.rvh.subbasins = subbasins
+        model.rvh.land_subbasins = land_group
+        model.rvh.lake_subbasins = land_group
+        model.rvh.reservoirs = reservoirs
+        model.rvh.hrus = hrus
```
huard, line 1168: Would returning a dict make things cleaner?
Something like 
`model.rvh.update(**rvh.importer.extract())`

No need to do it now, just a thought.

tests/test_emulators.py
```diff
@@ -1060,3 +1072,142 @@ def test_simple(self):
...
+            grid_weights=temperatures_importer.extract(),
+        )
+
+        model.rvt.gridded_forcings = [streaminputs_gf, temperatures_gf]
+
+        #############
+        # Run model #
+        #############
+
+        model([vic_streaminputs_nc_path, vic_temperatures_nc_path])
```
huard, line 1213: Can we check that the model output matches the tutorial output ?

tests/test_emulators.py
cjauvin: Yes absolutely, I thought about it too.

cjauvin:
@dustming Welcome to RavenPy! You will be interested in that PR I think.

In particular, it is making use of code that I adapted from your Routing Toolkit: https://github.com/CSHS-CWRA/RavenPy/blob/a27f6575600809ed74221eb580631a51fa236c63/ravenpy/models/importers.py#L24

So yours is the first "importer", and the second is @julemai 's grid weight generation script, which is also playing an important role in that PR:

https://github.com/CSHS-CWRA/RavenPy/blob/a27f6575600809ed74221eb580631a51fa236c63/ravenpy/models/importers.py#L221

tests/test_emulators.py
cjauvin: @julemai and @dustming Could you propose a good strategy for this?

dustming: > @julemai and @dustming Could you propose a good strategy for this?

Compare simulated hydrograph in Hydrographs.csv?

tests/test_emulators.py
cjauvin: thanks!

tests/test_emulators.py
cjauvin: compare with what?

ravenpy/models/commands.py
```diff
@@ -179,17 +204,395 @@ class GriddedForcingCommand(NamedTuple):
...
+    def to_rv(self):
+        return dedent(self.template).format(
+            basin_states_list="\n".join(map(str, self.basin_states.values()))
+        )
+
+
+@dataclass
+class SoilClassesCommand(RavenConfig):
+    @dataclass
+    class Record(RavenConfig):
```
cjauvin, line 499: @huard This is my new idea of having an internal `dataclass` to encapsulate the records that I was telling you about, instead of having a top-level class. I think it makes sense and is much cleaner, but I will wait for your input before updating the previous ones.

tests/test_emulators.py
cjauvin: done!

tests/test_emulators.py
cjauvin: now done!

cjauvin:
@huard @julemai @dustming From my perspective this PR is now complete and ready for final review. There's still the question of computing the average annual runoff dynamically but I'm not sure if we want to do it in this PR or not (Julie told me that she has code for it, somewhere).

cjauvin:
I forgot to add the important information that after having spent many hours with Julie trying to understand the differences between the Lievre results and those from my test (and fixing them, with a few tweaks), we are now completely confident that they are 100% identical.

huard (commented):
LGTM. Let's roll with this.

PR #34, opened 2021-01-19
Fix caching issue with ESP forecasts

huard:
* Name individual runs
* Added workdir argument
* Updated version to 1.7 in setup.py


ravenpy/utilities/forecasting.py
```diff
@@ -46,21 +46,22 @@ def get_raven_states(model, **kwds):
 
     """
     # Run the model and get the rvc file for future hotstart.
-    m = get_model(model)()
+    m = get_model(model)(workdir=workdir)
+    print("aha:", kwds)
```
cjauvin, line 50: This print should be removed.

setup.py
```diff
@@ -159,7 +159,7 @@ def run(self):
         "dev": dev_requirements,
     },
     url="https://github.com/CSHS-CWRA/ravenpy",
-    version="0.1.6",
+    version="0.1.7",
```
cjauvin, line 162: We use `bumpversion` to increment the version, which is set to modify 3 or 4 different files (including this one).

PR #39, opened 2021-01-27
Add functions supporting the routing shapefile

Zeitsperre:
This PR does the following:

* Places the imported libraries for several functions at the top of `gis.py` for visibility.
* Updates requirements in `setup.py` and `environment.yml` (synchronizing and adding missing libraries).
* Modifies several docstrings to be a bit clearer in `gis.py`
* Refactors existing HyrdoBASINS function to create  `get_location_wfs` and `get_feature_attributes_wfs` and creates two functions using them to be used by routing-shapefile WPS processes in RAVEN.
* Renames a test class to follow CamelCase convention.

This is all in satisfying parts of https://github.com/Ouranosinc/raven/issues/244.


ravenpy/utilities/gis.py
```diff
@@ -3,35 +3,146 @@
...
+    layer = f"{layer}{level}"
+
+    if coordinates is not None:
+        wfs = WebFeatureService(f"{GEO_URL}/wfs", version="1.1.0", timeout=30)
+        try:
+            resp = wfs.getfeature(
+                typename=layer, bbox=coordinates, srsname="urn:x-ogc:def:crs:EPSG:4326"
+            )
+        except Exception as e:
+            raise Exception(e)
```
cjauvin, line 76: I'm aware there must be a good reason for this but what is it again?

cjauvin (approved):

ravenpy/utilities/gis.py
```diff
@@ -3,35 +3,146 @@
...
 hybas_domains = {dom: hybas_dir / hybas_pat.format(dom) for dom in hybas_regions}
 
 
+def _get_location_wfs(
+    coordinates: Tuple[
+        Union[int, float, str],
+        Union[str, float, int],
+        Union[str, float, int],
+        Union[str, float, int],
+    ] = None,
```
huard, line 43: Why set a default value for coordinates if the function will fail without it ?

ravenpy/utilities/gis.py
```diff
@@ -3,35 +3,146 @@
...
+
+    Returns
+    -------
+    str
+      A GML-encoded vector feature.
+
+    """
+    layer = f"{layer}{level}"
+
+    if coordinates is not None:
```
huard, line 69: I think you can get rid of this if coordinates is a positional argument.

huard: Unless you'd want it to default to something in particular ?

ravenpy/utilities/gis.py
```diff
@@ -3,35 +3,146 @@
...
 hybas_dir = Path(__file__).parent.parent / "data" / "hydrobasins_domains"
 hybas_pat = "hybas_lake_{}_lev01_v1c.zip"
 
 # This could be inferred from existing files in hybas_dir
 hybas_regions = ["na", "ar"]
-
 hybas_domains = {dom: hybas_dir / hybas_pat.format(dom) for dom in hybas_regions}
 
 
+def _get_location_wfs(
```
huard, line 37: The name suggests this is a generic function, whereas it looks tied to a very specific dataset.

ravenpy/utilities/gis.py
```diff
@@ -3,35 +3,146 @@
...
+        except Exception as e:
+            raise Exception(e)
+    else:
+        raise NotImplementedError
+
+    data = resp.read()
+    return data
+
+
+def _get_feature_attributes_wfs(
```
huard, line 84: Same generic function name comment.

huard (commented):
First round of comments. Will test later.

Zeitsperre:
@huard Thanks for the comments. Please feel free to keep them coming. One thing I'd love to do is organize the functions in `gis.py` and `utils.py`. GIS-related functions are peppered in both, but fixing them will need to be done alongside raven processes at the same time. Once we have this PR done, I can move on to performing this in another PR with RAVEN changes synonymously.

ravenpy/utilities/gis.py
Zeitsperre: Not really in this instance since I don't add much here, and it could potentially hide useful exception messages. I coded this a long time ago, so it's good to be removed, I think.

setup.py
```diff
@@ -18,14 +18,23 @@
...
+    "pandas",
+    "pyproj",
     "rasterio",
     "rioxarray",
+    "shapely",
     "statsmodels",
     "xarray",
     "xclim[gis]>=0.23",
+    "requests",
     "wheel",
```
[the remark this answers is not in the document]
Zeitsperre: I can see that `gdal` was already present. Do you know if there are any functions (outside of the GIS-based ones I've integrated) that use GDAL? Should I move this as well to an `environment-gis.yml`?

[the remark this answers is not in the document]
huard: models/importers.py imports osgeo, but I suggest this goes into extra anyway.

Zeitsperre:
@huard I'm adding the final tests for `gis.py` tomorrow where it will be good to be merged. Many of the functions in `utils.py` have been fixed in https://github.com/Ouranosinc/raven/pull/335, and I believe it makes sense to port those to RavenPy once both these PRs are merged. With some work, we can probably hammer all this out by next week.

Zeitsperre:
@huard @cjauvin This is good to be merged. You'll notice a lot of `FIXME` and `TODO` notes; These will be dealt with when both this PR and https://github.com/Ouranosinc/raven/pull/335 are merged. I didn't want to touch code that has been touched in `raven`, hence why there's a lot of them. Thanks for your patience.

cjauvin:
I will have a deeper look later but does it fix the problems we had with RTD?

Zeitsperre:
No, the docs are still broken. Given that the changes here touch the requirements, causing Sphinx to complain, I can start adding those changes in here.

Zeitsperre:
For some reason, `mock` doesn't seem to register. I'm getting errors for geospatial libraries not being found despite listing them to be mocked. @tlvu do you have any advice?

Zeitsperre:
Figured out my issues (dataclasses is not in the Python3.6 builtins, I was forgetting to mock `osgeo`).

 I have a theory as to why the doc builds are giving me 849 warnings. I'm seeing some docstrings that start with a `:`, such as:
```
class SubBasinsCommand(NamedTuple):
    """:SubBasins command (RVH)."""
```
I think this is causing `sphinx` to freak out. Additionally, classes that are both published in `user_api.rst` and in the auto-generated `ravenpy.models.rst` are causing TONS of section header warnings. That type of fix might need another PR as that requires restructuring the docs layout, I imagine.

ravenpy/__init__.py
```diff
@@ -1,3 +1,7 @@
 """Top-level package for RavenPy."""
 
 from .__version__ import __author__, __email__, __version__  # noqa: F401
+
+
+class RavenPyDependencyError(BaseException):
+    pass
```
huard, line 7: What is the added value of this over ImportError ?

tests/test_utils.py
```diff
@@ -0,0 +1,337 @@
...
+    from shapely.geometry import GeometryCollection, shape
+
+    import ravenpy.utils as utils
+except (ModuleNotFoundError, RavenPyDependencyError):
+    utils = False
+
+from .common import test_data
+
+
+@pytest.mark.skipif(condition=utils is False, reason="GIS dependencies are needed.")
```
huard, line 21: https://docs.pytest.org/en/stable/reference.html#pytest-importorskip

ravenpy/utilities/regionalization.py
```diff
@@ -3,7 +3,6 @@
 """
 
 import logging
-from pathlib import Path
```
huard, line 6: Why remove this ?

huard (commented):
I'm getting failing tests. looks like a fiona error.

ravenpy/__init__.py
Zeitsperre: If RavenPy is going to be a building block for other components (like Raven-WPS), it helps to know that the errors are coming from RavenPy. This just makes it more visible. Can replace with ImportError if we want.

ravenpy/utilities/regionalization.py
Zeitsperre: eh? This must be a typo. My bad.

ravenpy/__init__.py
huard: I don't have a strong opinion. Is this a common pattern ? 
I found this: https://stackoverflow.com/questions/27361427/how-to-properly-deal-with-optional-features-in-python
which uses ImportError as well with a custom message.

Zeitsperre: It would be really easy to remove. I just don't want to see users installing the GIS dependencies one at a time (annoying). A custom error grabs attention, but so does `ImportError`

huard: Agreed. We want one clear message that says, please install RavenPy with such and such instructions for this to work. 
So let's keep it as is for now in order to move on.

Zeitsperre:
It looks as though the last hiccup I have is that the values expected for certain GIS operations are dependent on versions of GDAL. Very frustrating.

ravenpy/__init__.py
Zeitsperre: I've sadly removed the `RavenPyDependencyError` but the message is still present. Good enough,

PR #41, opened 2021-01-28
Fix caching bug in regionalization

huard:
A better solution would be to run the different simulations in parallel, but let's open another issue for this.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 History
 =======
 
+0.2.1
+-----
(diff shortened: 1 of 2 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 History
 =======
 
+0.2.1
+-----
+Fix xarray caching but in regionalization.
```
cjauvin, line 7: `but -> bug`

cjauvin (approved):

PR #48, opened 2021-01-29
Fix missing diagnostic file problem

### LINKED ISSUE #46: Faulty diagnostic file link
* RavenPy version: 0.2.1

### Description

`Raven.parse_results` logic is broken when there are no observations and thus no diagnostic file. This causes issues with the WPS server looking for files that don't exist.

huard:
Fix #46 
Related: https://github.com/Ouranosinc/raven/issues/197

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,7 @@ History
 -----
 * Set wcs.getCoverage timeout to 120 seconds.
 * Fix `Raven.parse_results` logic when no flow observations are present and no diagnostic file is created.
+* Fix ECCC test where input was cached and shadowed forecast input data
 
 0.2.1
 -----
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,7 +6,7 @@ History
 -----
 * Set wcs.getCoverage timeout to 120 seconds.
 * Fix `Raven.parse_results` logic when no flow observations are present and no diagnostic file is created.
-* Fix ECCC test where input was cached and shadowed forecast input data
+* Fix ECCC test where input was cached and shadowed forecast input data.
 
 0.2.1
 -----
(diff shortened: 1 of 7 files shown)


ravenpy/models/base.py
```diff
@@ -532,15 +532,17 @@ def parse_results(self, path=None, run_name=None):
...
                 if key != "diagnostics":
                     raise exc
-
-            fns.sort()
-            self.ind_outputs[key] = fns
-            self.outputs[key] = self._merge_output(fns, pattern[1:])
+                fns = None
 
         self.outputs["rv_config"] = self._merge_output(self.rvs, "rv.zip")
+        print(self.outputs)
```
cjauvin, line 545: Do we want this print?

HISTORY.rst
```diff
@@ -2,10 +2,15 @@
 History
 =======
 
+0.2.2
+-----
+* Set wcs.getCoverage timeout to 120 seconds.
+* Fix `Raven.parse_results` logic when no flow observations are present and no diagnostic file is created.
+* Fix ECCC test where input was cached and shadowed forecast input data
```
cjauvin, line 9: Nitpicking but can we stick with using or not using periods (`.`) at the end of items please.

tests/test_ECCC_forecast.py
```diff
@@ -39,18 +39,9 @@ def test_forecasting_GEPS(self, tmpdir):
...
-                fcst = forecasting.get_recent_ECCC_forecast(
-                    region_coll, climate_model="GEPS"
-                )
-
-        # write the forecast data to file
-        ts = f"{tmpdir}/fcstfile.nc"
-        fcst.to_netcdf(ts)
+        # Collect test forecast data for location and climate model.
+        fcst_fn = get_local_testdata("eccc_forecasts/geps_watershed.nc")
+        fcst = xr.open_dataset(fcst_fn)
```
cjauvin, line 44: I would suggest to rename `fcst_fn` to `ts20`, to put emphasis on the fact that it's a 20 channel time series (as opposed to Salmon, used above) and I would get rid ot the opened dataset (`fcst`) entirely since it's only used to set the `nc_index` below in the model call. I think 20 should be simply set as a variable (`n_stations = 20` or something) to be used in the asserts.

cjauvin (approved):

ravenpy/models/base.py
huard: oups

HISTORY.rst
huard: Added periods everywhere.

tests/test_ECCC_forecast.py
huard: ok

tests/test_ECCC_forecast.py
```diff
@@ -62,7 +53,7 @@ def test_forecasting_GEPS(self, tmpdir):
         model.rvc.parse(rvc.read_text())
 
         model(
-            ts=(ts,),
+            ts=(fcst_fn,),
```
cjauvin: Also I'm not sure this param really needs to be wrapped in a 1-tuple.

PR #49, opened 2021-01-29
Bugfix: point to RavenPy/data and include it

Zeitsperre:
A pip install of `RavenPy` was not installing the necessary HydroBASINS helper shapefiles. This fix ensures that the folder is being pointed to properly in the MANIFEST.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,5 +12,5 @@ recursive-exclude * *.py[co]
 
 recursive-include docs *.rst conf.py Makefile make.bat *.jpg *.png *.gif
 
-graft data
+graft ravenpy/data
 graft ravenpy/models


cjauvin:
really?? I thought I had checked that many times..

cjauvin:
I even made experiments with that: https://github.com/CSHS-CWRA/RavenPy/blob/master/setup.py#L146

cjauvin:
Actually I was curious and I reinstalled ravenpy in a fresh conda env, and the data files seem to be there:

```shell
(test-ravenpy) ~/anaconda3/envs/test-ravenpy$ ll lib/python3.8/site-packages/ravenpy/data/hydrobasins_domains 
total 4.4M
drwxrwxr-x 2 christian christian 4.0K Jan 29 18:33 .
drwxrwxr-x 3 christian christian 4.0K Jan 29 18:33 ..
-rw-rw-r-- 1 christian christian 2.1M Jan 29 18:33 hybas_lake_ar_lev01_v1c.zip
-rw-rw-r-- 1 christian christian 2.4M Jan 29 18:33 hybas_lake_na_lev01_v1c.zip
```

Zeitsperre:
@cjauvin I tried making some source packages and wheel builds of it (`python setupy.py sdist bdist_wheel`) and I wasn't seeing them in the wheel files for some reason. This might be due to the fact that using `install-options`, doesn't create a wheel (no harm done), but the data files should at least be present in both distributions (source, wheel). This is just a fix for something that was tripping me up on Friday.

cjauvin:
@Zeitsperre I'm looking right now at the 0.2.2 wheel and sdist I produced last week (and that I sent to PyPI) and `ravenpy/data/hydrobasins_domains/` is present in both of them. There seems to be something weird on your end. Are you installing with `pip` or `python setup.py`?

Zeitsperre:
I was using `pip`, but I was upgrading from an existing installation. If this isn't an issue, perhaps it's something on my configurations. In any case, if it's working, let's not break it.

cjauvin:
I mean your fix is really not a big deal and the syntax seems quite logical, but this is a case where I would actually prefer to get a fine-grained understanding of what goes on, and how it works.

Zeitsperre:
This fix has been integrated into #52

PR #59, opened 2021-02-10
Reference -> References in docstring, to make Sphinx happy

cjauvin:
We have broken deps also.


Zeitsperre (approved):
This one sparks joy.

PR #60, opened 2021-02-10
Aggregate nc based on gridweights

julemai:
This is a working version. Certainly needs to be fancy'd up to fulfill standards. A test is added to check revised weights and the new NetCDF file


huard:
Hi @julemai 

Could add a high-level explanation of what this does and why it's useful ?

julemai:
Hi @huard 

Yes, for sure. Sorry that I didn't do this earlier.

The purpose is to reduce input NetCDF files usually in (lat,lon,time) dimension to (nHRUs,time) where the HRUs being the actual processing units in Raven. This aggregation is done exactly why we derive the grid-weights and Raven does that usually at first after it read in the NetCDF file. 

We however realized that if the original NetCDF file is huge (for example a global ERA5 file), Raven is spending huge amounts of the runtime with only reading the NetCDF files and gathering the few cells that are relevant for the HRUs. I am thinking that this is the exact reason why Richard is reporting huge runtimes when he is trying to calibrate some models. 

So, the idea is to reduce the size of the NetCDF files and do this HRU aggregation outside of Raven- especially for calibration setups. I wouldn't bother with just a single forward model run.

Does that make sense?

huard:
Totally !

Would it make sense to refactor the function in three pieces ?

1. Convert existing weight file to ilat, ilon, weights, hru_id arrays. 
2. Flatten netCDF file according to output from previous function
3. Write new flat weight file from previous output

huard:
My angle is that we've worked on https://pangeo-xesmf.readthedocs.io/en/latest/notebooks/Spatial_Averaging.html over the last months, and my guess is that we could use it to compute the weights and average over the HRUs. These weights are saved in a netCDF that can be read as a sparse array. https://github.com/pangeo-data/xESMF/blob/9177e3000bc808cd166382525772b2472b3cb3da/xesmf/smm.py#L13

It would be nice to have interoperability between your code and format and the xesmf functionality. I'm not yet convinced that xESMF would be significantly faster than your code, but this is something I'd like to benchmark during our maintenance phase.

julemai:
Re refactoring into two separate tasks:
- The weight file already contains exactly this information. The only change would be cell_id --> ilon,ilat
- I'm fine with that if that makes more sense to you

I'm totally game of using something which is faster. I just usually use GDAL functions directly as they are usually the fastest option. But if the results are the same I am totally for it if you find something faster. We should just benchmark with something larger than the current test-data used here.

huard:
I'm not suggesting we do this right now. Let's use the code that you wrote that we know works. I'm not even sure xesmf would be faster anyway.

julemai:
ok. Benchmarking later...

Re refactoring:
I will add a new function `weights-cell-id-to-lat-lon` and then use that one for flattening.

julemai:
Refactored into new separate function but not really sure if that makes it easier

ravenpy/models/importers.py
```diff
@@ -283,27 +283,28 @@ def extract(self) -> GridWeightsCommand:
...
+            hru = g[hru_id][:1]
+            did = g[g["DowSubId"] != -1]["DowSubId"].values[0]
+            onm = g[g["Obs_NM"] != -9999]["Obs_NM"].values[0]
+            other_cols = g[[c for c in g.columns if c not in [hru_id, "DowSubId", "Obs_NM"]]]
+            # Make sure that all the other values are the same for every row
+            assert other_cols.value_counts().shape == (1,)
+            return pd.DataFrame(dict({hru_id: hru, "DowSubId": did, "Obs_NM": onm}, **other_cols[:1]))
+
+        # Remove duplicate HRU_IDs while making sure that we keed relevant DowSubId and Obs_NM values
+        self._routing_data = self._routing_data.groupby(self._routing_id_field).apply(keep_only_valid_downsubid_and_obs_nm)
```
cjauvin, line 307: @julemai I think this will work but the problem is that it not being tested, as we have no test data where the special aggregation logic is needed. The ideal would be to introduce one such dataset in `raven-testdata` (and a corresponding test here), to avoid bad surprises later.

julemai: Yeah, I was worried that you would ask for that. :) I'll put it on the list.

ravenpy/models/aggregate_forcings_to_HRUs.py
```diff
@@ -0,0 +1,232 @@
...
+import netCDF4 as nc4
+import numpy as np
+
+
+DIM_NAMES = ("lon", "lat")
+VAR_NAMES = ("lon", "lat")
+ROUTING_ID_FIELD = "HRU_ID"
+NETCDF_INPUT_FIELD = "NetCDF_col"
+AREA_ERROR_THRESHOLD = 0.05
+VARIABLES_TO_AGGREGATE = ["Streaminputs"]
```
huard, line 14: That's an unexpected default value from my POV. I suspect that in the majority of cases, users will aggregate forcing fields (pr, tas). I suggest that the default variable are left as dynamic and determined by the variables that are defined over dimensions that are a subset of DIM_NAMES. 
See https://github.com/pangeo-data/xESMF/pull/79/files for an example of this logic in action.

ravenpy/models/aggregate_forcings_to_HRUs.py
```diff
@@ -0,0 +1,232 @@
...
+       latitude ID (ilat) and longitude ID (ilon);
+       both starting with 0
+
+    Return:
+      gridweights structure with (ilon, ilat) instead of (cell_id)
+
+    Arguments:
+      gws  (structure) ... grid-weights-structure as returned by RoutingProductGridWeightImporter and then extracted()
+      nlon (integer)   ... number longitudes NetCDF file
+      nlat (integer)   ... number latitudes  NetCDF file
```
huard, line 29: We're using https://numpydoc.readthedocs.io/en/latest/format.html for the docstring convention.

ravenpy/models/aggregate_forcings_to_HRUs.py
```diff
@@ -0,0 +1,232 @@
...
+# routing_file = "/Users/j6mai/Documents/GitHub/GridWeightsGenerator/example/maps/finalcat_hru_info.zip"
+# output_file = "/Users/j6mai/Documents/GitHub/GridWeightsGenerator/example/input_VIC/VIC_streaminputs_agg.nc"	
+# variables_to_aggregate=VARIABLES_TO_AGGREGATE
+# dim_names=DIM_NAMES
+# var_names=VAR_NAMES
+# routing_id_field=ROUTING_ID_FIELD
+# netcdf_input_field=NETCDF_INPUT_FIELD
+# gauge_ids=None
+# sub_ids=None
+# area_error_threshold=AREA_ERROR_THRESHOLD
```
huard, line 229: remove

ravenpy/models/base.py
```diff
@@ -398,16 +398,14 @@ def run(self, ts, overwrite=False, **kwds):
...
-            ]
+            ts = [ts]
 
         # Case for potentially parallel parameters
         pdict = {}
         for p in self._parallel_parameters:
             a = kwds.pop(p, None)
 
-            if a is not None and p in ["params", "basin_state", "hru_state"]:
+            if a is not None and p in ["params"]:
```
huard, line 408: Won't this break support for parallel runs on multiple hru_states ? This functionality going to be needed for assimilation.

huard (approved):
Could only do a superficial review, but it looks good. Thanks for all the additional comments, really useful.

ravenpy/models/base.py
huard: I realize this is coming from another branch. Please ignore.

cjauvin: Not sure where this is coming from, this PR is bound to be very complex (I'm merging my routing PR to avoid too many future conflicts)

julemai (commented):
Looks excellent to me! Thanks @cjauvin and @huard

PR #62, opened 2021-02-16
Allow installing binaries in develop/editable mode

cjauvin:
This should fix a pain point I had for a while: the inability to use `--install-option="--with-binaries"` in editable (aka. develop) mode. The overall solution is still not perfect of course, but at least this should make it easier for the most common scenario our users want: 
1. git clone
2. pip install in editable mode
3. getting the binaries.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,8 @@ import urllib.request
 import zipfile
 from pathlib import Path
 
-from setuptools import find_packages, setup
+from setuptools import Distribution, find_packages, setup
+from setuptools.command.develop import develop
 from setuptools.command.install import install
 
 with open("README.rst") as readme_file:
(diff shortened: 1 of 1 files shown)


Zeitsperre:
This doesn't seem to be working in a clean environment. GIS libraries attempt to build wheels but using `install-option` aborts the wheel-making processes:

image

Should I be installing without binaries (to build wheels) then running the same install command `--with-binaries` to get the RAVEN/OSTRICH files? If so, that needs to be clearer.

cjauvin:
I'm not sure what goes on but I don't think the errors you get are directly related to this PR. It's always better to conda install all that is listed in environment.yaml, and in the end, as a separate step, to pip install -e your codebase, with which you can now fetch the binaries as an extra bonus. I think --install-option always disables wheel building, editable or not.

cjauvin:
The way I see it, this part of the docs https://ravenpy.readthedocs.io/en/latest/installation.html#from-sources-for-development-and-testing is now easier because you can get the binairies at the same time you pip install.

Zeitsperre:
Ok it seems to be working. My terminal was stuck at `Running setup.py develop for ravenpy` for a few minutes before it succeeded:

image

It would be good to have some output to the terminal, just so I can know if the binaries are being compiled/installed. I'm guessing that the `RAVENPY_TESTDATA_PATH` still also needs to be set at some point for tests?

cjauvin:
You can use `--verbose` I think for a little more info. The test data could be handled in a similar way, but the problem is different than with the binaries: where exactly would the installer put them? It's less clear, and also I think it's easier if the test data has been git cloned, for easier refresh. All that is totally open for debate of course, I am just exploring the space of possible solutions.

Zeitsperre (approved):

tlvu:
So with this fix, on raven-wps side no need to uninstall RavenPy and re-install with `--with-binaries` to get the binaries? See https://github.com/Ouranosinc/raven/blob/cb15c50f56255c80b811b082cab8b8e80bee7d4f/Makefile#L104-L110 and https://github.com/Ouranosinc/raven/blob/cb15c50f56255c80b811b082cab8b8e80bee7d4f/Dockerfile#L27 Would be pretty cool to not have to un-install and re-install !

Does it still invoke `make` as previous `--with-binaries`?  If yes, then maybe have to specify in the docs somewhere it also needs `make`, `g++`, `libnetcdf-dev` as discovered in https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/blob/c3da54190ecd1f6f7b7a84fe5eb02e1599c81512/docker/Dockerfile#L35-L40 when trying to add RavenPy to Jupyter env.

Does it replace `--with-binaries` all together?

cjauvin:
@tlvu Sorry for not being more clear: this fix is JUST to allow `pip install -e <path> --install-option="--with-binaries"` (i.e. editable AND with-binaries), which currently is not working (only NON-editable works with `--install-option="--with-binaries"`).

Moreover you are right: yes this STILL invokes `make` and `g++`, so `sudo apt-get install build-essentials libnetcdf-dev` is STILL needed in any case, and should be better documented. 

More moreover: before pip installing ravenpy (editable or not) it's actually better to have conda install all its deps, because of GDAL in particular, which always insists on being such a nasty brat (conda install beats him into submission well it seems, better than pip).

cjauvin:
Again this fix is not meant to be the final solution, but just a way to avoid having to do something silly like:

1. `pip install ravenpy --install-option="--with-binaries"`
2. `pip uninstall ravenpy`
3. `pip install -e .`

In order to get the binaries and place them where they need to be!

cjauvin:
I think this fix is mostly relevant and nice for dev setup, and not really for infra stuff.

cjauvin:
> So with this fix, on raven-wps side no need to uninstall RavenPy and re-install with --with-binaries to get the binaries?

No actually this is rather due to the fact that this line is present: 

https://github.com/Ouranosinc/raven/blob/master/environment.yml#L48

So conda installs it itself, but I don't think it could deal with the `--install-option="--with-option"` in the yaml, which is a pip/setuptools specific option.

I think the best would be to remove ravenpy from `environment.yaml`, and install it with pure pip.

tlvu:
> @tlvu Sorry for not being more clear: this fix is JUST to allow `pip install -e <path> --install-option="--with-binaries"` (i.e. editable AND with-binaries), which currently is not working (only NON-editable works with `--install-option="--with-binaries"`).

No worries.  I was just carried away with excitement.

> I think the best would be to remove ravenpy from `environment.yaml`, and install it with pure pip.

The current kludge to uninstall and reinstall with `--with-binaries` already works.  I wouldn't put effort on another work-around to fix an existing work-around.  I would focus on fixing the real root cause that forced us to use the work-around in the first place: there is no proper way to distribute raven binaries !  To this end, a proper Conda or/and Pip package is the answer.  I wonder if we need both.

cjauvin:
@tlvu Again this is not meant to be the final ultimate solution but it DOES solve a real and annoying problem for RavenPy users, until we have better. Not everyone lives in a docker container :wink:

cjauvin:
That said I totally understand your point, which is very valid!

tlvu (approved):

PR #66, opened 2021-03-01
Add climpred verification

richardarsenault:
Added climpred verification capabilities to ravenpy. Basically, for a series of hindcast dates, we can now compute reliability diagrams, rank histograms, CRPS and all the other probabilistic metrics used in climpred/xskillscore.

I've had to add code to forecasting.py utilities and I've created a test function which passes on my end for 3 hindcast verification metrics. 

Notes:
1- It is kind of slow, Raven must be run multiple times to generate the hindcasting data. However I assume that anyone somewhat competent will find ways to optimize the code :)

2- I've had to install climpred and a few dependencies which are not currently in the environment, so I think that will have to be added somehow.

All comments/suggestions are welcome!


huard (commented):
I don't understand why evaluating the forecast verification requires running the model ? Why can't we just call `perform_climatology_esp` and then analyze the results ? I'm missing something.

richardarsenault:
Yes, that is what we are doing here. Perform_climatology_esp is called in the function. However, perform_climatology_esp needs to run raven to be able to produce the ESPs. So I am passing them along the chain the entire time until perform_climatology_esp runs the model and generates the ESPs for each hindcast date.

Zeitsperre:
Just making a note here that this PR is dependent on #6 being merged ahead of it.

huard:
You lost me here. `perform_climatology_esp` runs the model already, no ?

tests/test_climatology_ESP_verfication.py
```diff
@@ -0,0 +1,64 @@
...
+
+@author: ets
+"""
+
+import datetime as dt
+
+from ravenpy.utilities.testdata import get_local_testdata
+from ravenpy.utilities.forecasting import make_ESP_hindcast_dataset, make_climpred_hindcast_object
+
+class TestClimpredHindcastVerification:
```
huard, line 14: This test takes about 65 seconds to run on my end. I suggest we use the minimum forecast duration and number of years to get a valid test (in terms of software, not hydrology) to speed up the test.

huard (approved):
Happy with this PR after test speed-up (if possible). I suggest we defer optimizations (parallelism) to the maintenance phase. Will open a ticket.

tests/test_climatology_ESP_verfication.py
richardarsenault: I've shortened the period and forecast duration so that it is a super small test, takes 11 seconds now on my VM. Can't go faster unless we change the ESP dataset, and as far as I know we don't have another one like this lying around (as we use all years present in the dataset for now).

Agreed with all the rest (i.e. optimizations in the maintenance phase).

Can we merge now or is there still a problem with versioning as mentioned by @Zeitsperre with #6?

Zeitsperre: The issue is that, as I understand it, `climpred` support is dependent on #6. If that isn't the case, I can add it to the setup requirements on this branch.

richardarsenault: OK! @huard can confirm but I did not change anything nor pull the #6 PR branch when I did my setup, so I think it might work anyways?

Zeitsperre: Alright. In that case, I can port my changes from there to here.

Zeitsperre: Done!

PR #70, opened 2021-03-04
Fix the Docs

Zeitsperre:


Zeitsperre:
This PR still doesn't fix the issues. I've reduced a lot of circular imports but all the docstrings under `models` are malformed. Will return to this tomorrow.

cjauvin:
I am trying to push this PR further but I find it very hard, as I don't know much about the tool stack in use and they seem extremely finicky. So after created a new conda env with the `environment-rtd.yaml` file, the first problem I have is with `__version__`, which is apparently defined twice. The next problem I have is that if this is commented out (as it is, no idea why that is the case):

https://github.com/CSHS-CWRA/RavenPy/blob/fb14e6c930f383863a4760d2f4d616bddea8cf78/docs/conf.py#L39

then I have errors with "unexpected section titles", with docstring headers such as:

https://github.com/CSHS-CWRA/RavenPy/blob/fb14e6c930f383863a4760d2f4d616bddea8cf78/ravenpy/models/base.py#L47

After uncommenting this extension, the next error I get is: ```Explicit markup ends without a blank line; unexpected unindent.```

and after reading this https://stackoverflow.com/a/38762633/787842 after a quick google search, I realize that I don't want to spend my entire evening with a tool which becomes unable to compute the moment it is not happy with the exact number of white lines and space characters that I have in my docstrings, and that's where I decide to give up (for tonight).

cjauvin:
Oops I just realized that that's probably exactly what you meant by:

> but all the docstrings under models are malformed

:pleading_face:

PR #71, opened 2021-03-08
Add routing support to emulators

cjauvin:
Add routing support into GR4JCN emulator. 

I'm adding it to not forget:
- [x] Implement average annual runoff (not sure if we want to do it in the context of this PR or not?)


ravenpy/models/commands.py
```diff
@@ -563,19 +563,23 @@ def to_rv(self):
...
+        profile_name: str = ""
+        soil_class_names: Tuple[str] = ()
+        thicknesses: Tuple[float] = ()
 
         def to_rv(self):
-            return " ".join(map(str, asdict(self).values()))
+            # From the Raven manual: {profile_name,#horizons,{soil_class_name,thick.}x{#horizons}}x[NP]
+            n_horizons = len(self.soil_class_names)
+            horizon_data = [
+                a for b in zip(self.soil_class_names, self.thicknesses) for a in b
```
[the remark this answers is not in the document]
cjauvin: done.

huard:
Hi @julemai 

It looks like the GR4JCN emulator now can be run with routing! 

I've followed your instructions to modify the model templates, and created a dummy test exercising two subbasins with routing from one to the other. See `tests/test_emulators.py::TestGR4JCN::test_routing`

Could you check that 
1. The changes in the templates are OK 
2. The test makes correct assumptions and indeed tests the routing

Note that I've hardcoded in the test the output directory (`model = GR4JCN("/tmp/test_gr4jcn_routing")`) so when you've run the test you can easily go into the directory and check the config files.

julemai:
Hi @huard 

Yay!! That's fantastic. I will get to that first thing Monday morning. Once it passes my review I might need to share it with James to get his feedback too.

Thanks so much for fixing that output directory. That will speed up my testing by a factor of 10!!

julemai:
Updated above with a code snippet for `Annual_Avg_Runoff `...

cjauvin:
> Fixing it for now is fine I think.

So for this current phase (until the end of the month) are we using a fixed value or do we try to compute it with the method you outlined?

julemai:
@cjauvin I think `avg_runoff` is an easy fix since everything (Qobs and mean area) is already available as far as I know. But up to you and @huard. I'm good with either.

The more severe thing to look is it the RVT. We need a fake grid-weights looking like this (we are going to use the same forcings for both HRU; just because the inputs right now only contain one time series; this will change once we don't work with fake examples...):
```
:GridWeights
	:NumberHRUs 2
	:NumberGridCells 1
	1 0 1.0
	2 0 1.0
:EndGridWeights
```

And the block for reading the inputs will need to look like this (but for this we need to add a new axis (`nstations`) to the existing NetCDF:
```
:StationForcing rain
  :ForcingType     rain
  :FileNameNC      Salmon-River-Near-Prince-George_meteo_daily_2D.nc
  :VarNameNC       rain
  :DimNamesNC      nstations ntime
  :RainCorrection  1.0 
  :SnowCorrection  1.0
  :tshift          0.0
  :LinearTransform 1.0 0.0
  :RedirectToFile  <grid-weights-from-above>
:EndStationForcing
```

I think we should use `StationForcings` for ALL examples. The only thing we need is to add this additional axis. The rest would be the same. It would make it more consistent and general. Makes sense?

cjauvin:
So we need a variation on `Salmon-River-Near-Prince-George_meteo_daily_2D.nc` with a new axis? Could I ask you to create it please, and then I can add it to `test-ravendata` ?

julemai:
Sure. I can do that. Some time id week. Before I will be busy with integrating the other two models.

julemai:
Already exists:
```
testdata/raven-gr4j-cemaneige/Salmon-River-Near-Prince-George_meteo_daily_2d.nc
```

The axis is called there `region` and not `nstations`. So, you would need to adjust that in what I exemplarily had above.

julemai:
These 2D NetCDF files and the needed grid-weights are btw exactly the ones that are the result of `aggregate_forcings_to_hrus.py`. So, it would be really convenient to just use those here. :)

cjauvin:
@julemai and @huard My latest commit introduces (hopefully!) the missing pieces to make the `TestGR4JCN.test_routing` run (as advised by Julie) although I'm not sure it's all correct, please verify (in `/tmp/test_GR4JCN_simple` after running the test).

julemai:
That’s awesome! I will have a look tomorrow.

tests/test_emulators.py
```diff
@@ -92,18 +100,59 @@ def test_simple(self):
...
+        np.testing.assert_almost_equal(d["DIAG_NASH_SUTCLIFFE"], -0.116971, 4)
 
+        # ------------
+        # Check simulated streamflow values q_sim
+        # ------------
         hds = model.q_sim
+
         assert hds.attrs["long_name"] == "Simulated outflows"
 
+        assert len(hds.nbasins == 2)  # ??? WHY IS THAT 2??
```
julemai, line 116: Need to check why there is two basins in this output and not only one.

tests/test_emulators.py
```diff
@@ -92,18 +100,59 @@ def test_simple(self):
...
+        # Check parser
+        # ------------
+        assert model.rvi.calendar == "GREGORIAN"
+
+        # ------------
+        # Check saved HRU states saved in RVC
+        # ------------
+        assert 1 in model.solution["HRUStateVariableTable"]["data"]
+
+        # ------------
```
julemai, line 153: Just added the same checks as we do also with `test_routing` for some consistency

tests/test_emulators.py
```diff
@@ -128,27 +177,31 @@ def test_routing(self):
         #########
 
         # In this test the HRU IDs range from 0 to 2, and the subbasin IDs are 10, 20
-        model.rvc.basin_state.index = 10
+        model.rvc.basin_state.index = 10  # <<<<<<< THIS NEEDS TO BE TWO INSTANCES
+        model.rvc.basin_state.index = 20  # <<<<<<< THIS NEEDS TO BE TWO INSTANCES
```
julemai, line 181: ToDo. We need to create two of these blocks in the RVC file

tests/test_emulators.py
```diff
@@ -197,8 +250,21 @@ def test_routing(self):
         model.rvp.params = model.params(0.529, -3.396, 407.29, 1.072, 16.9, 0.947)
 
         # TODO: compute this??
+
+        # model.rvp.avg_annual_runoff = function_to_derive_this(area_as_sum_of_all_hrus, q_obs_most_down_stream_means_with_down_sub_id=-1)
```
julemai, line 254: ToDo: Derive average annual runoff based on total basin area and `q_obs` of the most downstream basin (the one with `DownSubId==-1`)

ravenpy/models/emulators.py
```diff
@@ -114,13 +114,17 @@ def derived_parameters(self):
             soil0 = self.rvd.GR4J_X1_hlf if self.rvc.soil0 is None else self.rvc.soil0
             soil1 = self.rvc.soil1
 
+        # subbassin_id -> has at least one LakeHRU
+        sb_is_lake = defaultdict(lambda: False)
```
julemai, line 118: Better name might be `sb_contains_lake`.

ravenpy/models/emulators.py
```diff
@@ -131,6 +135,17 @@ def derived_parameters(self):
...
+            [sb.subbasin_id for sb in self.rvh.subbasins if sb_is_lake[sb.subbasin_id]]
+        )
+        self.rvh.land_subbasins = tuple(
+            [
+                sb.subbasin_id
+                for sb in self.rvh.subbasins
+                if not sb_is_lake[sb.subbasin_id]
+            ]
+        )
+
```
julemai, line 148: That's pretty awesome!!

Could you please then also add the following two lines below these `SubBasinGroup`'s?
```
:SBGroupPropertyMultiplier	Land	MANNINGS_N	            1.0
:SBGroupPropertyMultiplier	Lakes	RESERVOIR_CREST_WIDTH	1.0
```
These lines are the actual reason why we create those groups. Otherwise the groups are not used in the setup. Plus we will need these two lines when we start calibrating the GR4J+routing example. In calibration mode the two `1.0`'s will be converted into two parameters (e.g., `par_x08` and `par_x09`). Merci.

cjauvin:
@julemai Yes absolutely, that's on my todo list but yesterday night I had to stop just before it. So the current status of this PR is:
- [ ] Set HRU states to 0
- [ ] Have 1 BasinIndex per SB
- [ ] Set IDs in SB land/lake groups
- [ ] Add SB Group properties
- [ ] Compute avg annual runoff
- [ ] Go back and fix all the other tests that are now broken

julemai:
Sounds good. I saw that you were on a roll yesterday night. :)

ravenpy/models/emulators.py
cjauvin: You greatly activated my special fiber for "variables should have semantically accurate names", thanks and done!

ravenpy/models/emulators.py
cjauvin: done!

tests/test_emulators.py
cjauvin: done

cjauvin:
@julemai @huard A lumped GR4JCN model must now have a valid (ie. non-zero area) initial HRU defined:

https://github.com/CSHS-CWRA/RavenPy/blob/f7964ff9becb61887d155809f2912d47f9715919/tests/test_emulators.py#L407

which means that just doing this in the emu ctor is not enough:

https://github.com/CSHS-CWRA/RavenPy/blob/f7964ff9becb61887d155809f2912d47f9715919/ravenpy/models/emulators.py#L98

(because if not Raven complains that the HRU has zero or negative size).

So my question: now that we are being explicit about it (because we have generalized the lumped/not lumped use cases) should we modify the emulator constructor somehow, to make it more obvious?

julemai:
Hi @cjauvin! I am confused. The HRU _always_ had a valid area and it is 4650.6 km2. The areas will be given in the routing product of by another external source that is delineating the basin. The simple test (lumped model) was definitely running when I tested the routing test. I always did both... confused

cjauvin:
Ah you are right, I had not realized that we had that before:

https://github.com/CSHS-CWRA/RavenPy/blob/master/tests/test_emulators.py#L70-L74

and that:

https://github.com/CSHS-CWRA/RavenPy/blob/master/tests/test_emulators.py#L127-L130

cjauvin:
`TestGR4JCN.test_resume` fails:
```
        for key in ["Soil Water[0]", "Soil Water[1]"]:
>           np.testing.assert_array_almost_equal(
                model_a.storage[key] - model_ab.storage[key], 0, 5
            )
E           AssertionError:
E           Arrays are not almost equal to 5 decimals
E
E           Mismatched elements: 184 / 185 (99.5%)
E           Max absolute difference: 0.00228928
E           Max relative difference: inf
E            x: array([ 7.79141e-07, -3.16415e-04, -5.87290e-04, -5.85631e-04,
E                  -8.07810e-04, -1.00260e-03, -1.18140e-03, -1.33439e-03,
E                  -1.46652e-03, -1.58590e-03, -1.70162e-03, -1.79083e-03,...
E            y: array(0)
```

The diference is likely related to the way we now parse the RVC but it eludes me.. I've been looking at before/after values and the parsing seems ok to me.. could you please have a look?

julemai:
If the setup files are EXACTLY the same as before the results will be EXACTLY the same. Could you compare the setup files (all 5 RV* files) from before and after and let me know what is different?

cjauvin:
This is the "resume" scenario where a model runs in different steps (after
dumping and rereading its state in a solution.rvc); I tried to do what you
say but this setup is much harder for me to analyze. But anyway I will try
to continue my debugging using other comparison techniques.

tests/test_emulators.py
```diff
@@ -133,11 +398,14 @@ def test_run(self):
         )
         d = model.diagnostics
 
-        np.testing.assert_almost_equal(d["DIAG_NASH_SUTCLIFFE"], -0.117301, 2)
+        np.testing.assert_almost_equal(d["DIAG_NASH_SUTCLIFFE"], -0.116971, 2)
```
cjauvin, line 401: TODO: check this one!

cjauvin: This difference was due to lat/lon values of 0 in the `:Gauge` of the RVT.

julemai: ok. Great. Maybe we should then test for 4 digits and not just 2 to catch this...

tests/test_emulators.py
julemai: Fixed. Needs to be `assert (len(hds.nbasins) == 1)`

tests/test_emulators.py
julemai: Using now same reference `"DIAG_NASH_SUTCLIFFE"` everywhere in code (i.e. `-0.116971`) even though it is the wrong one. The `Latitude`, `Longitude`, and `Elevation` of the entire basin needs to be passed on to `Gauge` as it was done before using:

```
kwargs = dict(
            area=4250.6,
            elevation=843.0,
            latitude=54.4848,
            longitude=-123.3659,
            params=(0.529, -3.396, 407.29, 1.072, 16.9, 0.947),
        )
```

These values are for the entire BASIN and not for a sub-basin or HRU... After that all `"DIAG_NASH_SUTCLIFFE"` should be `-0.117301`.

tests/test_emulators.py
```diff
@@ -318,7 +318,7 @@ def test_routing(self):
 
         hds = model.q_sim
 
-        assert len(hds.nbasins == 2)
+        assert len(hds.nbasins) == 1  # number of "gauged" basins is 1
```
cjauvin, line 321: :grin:

huard:
Ok to merge ?

PR #77, opened 2021-03-12
Adding model SAC-SMA

julemai:
All tests passing.

Details about SAC-SMA can be found here:
https://wiki.ewater.org.au/display/SD41/Sacramento+Model+-+SRG


julemai:
This needs a `Raven-rev311` executable to work...

ravenpy/models/emulators.py
```diff
@@ -640,6 +646,158 @@ def ost2raven(self, ops):
...
+            POW_X02=None,
+            POW_X03=None,
+        )
+
+    def derived_parameters(self):
+        self.rvd["par_soil0_mm"] = self.rvp.params.par_x04 * 1000.0
+        self.rvd["par_soil2_mm"] = self.rvp.params.par_x06 * 1000.0
+        self.rvd[
+            "PAR_BF_LOSS_FRAC"
+        ] = self.rvp.params.par_x12  # PAR_BF_LOSS_FRAC = par_x12/(1+par_x12)
```
cjauvin, line 692: Please put your comments on the previous line (not at the end of the line) so that Black does not do this strange formatting (due to line lenght being > 88). Also if you are not already doing it, please use `pre-commit`, in your venv:
```
pip install pre-commit # if not already installed in your venv
pre-commit install
```
This will run some formatting and verification tools on your source code (black, isort and flake8) just before the commit stage, and it will prevent the commit from happening if it cannot resolve all the problems by itself.

cjauvin (approved):

julemai:
I have done all the formatting that @cjauvin wanted me to include. Tests are all passing.

ravenpy/models/emulators.py
```diff
@@ -641,6 +663,160 @@ def ost2raven(self, ops):
...
+        self.rvd["par_soil0_mm"] = self.rvp.params.par_x04 * 1000.0
+        self.rvd["par_soil2_mm"] = self.rvp.params.par_x06 * 1000.0
+        # PAR_BF_LOSS_FRAC = par_x12/(1+par_x12)
+        self.rvd["PAR_BF_LOSS_FRAC"] = self.rvp.params.par_x12
+        # 10.0**self.rvp.params.par_x01  #
+        self.rvd["POW_X01"] = self.rvp.params.par_x01
+        # 10.0**self.rvp.params.par_x02  #
+        self.rvd["POW_X02"] = self.rvp.params.par_x02
+        # 10.0**self.rvp.params.par_x03  #
+        self.rvd["POW_X03"] = self.rvp.params.par_x03
```
huard, line 719: Why do you need those ?

huard (approved):
Looks all good.

ravenpy/models/emulators.py
julemai: These are derived parameters. They will have a different name in the Ostrich setup and we probably want them to be named the same- no matter if it is in a Raven run or Raven run within the Ostrich calibration setup. Derived parameters exist in ALL other setups too and I tried to follow the logic we are using there too.

ravenpy/models/emulators.py
julemai: Done.

ravenpy/models/emulators.py
```diff
@@ -630,9 +650,11 @@ def ost2raven(self, ops):
         BLENDEDParams named tuple
           Parameters expected by Raven.
         """
-        names = ["par_x{:02}".format(i) for i in range(1, 36)]+["par_r{:02}".format(i) for i in range(1, 9)]
-        names[3]  = "pow_x04"
-        names[9]  = "sum_x09_x10"
+        names = ["par_x{:02}".format(i) for i in range(1, 36)] + [
+            "par_r{:02}".format(i) for i in range(1, 9)
```
cjauvin, line 654: You can use an f-string instead of `format` here: `f"par_x{i:02}"`

ravenpy/models/emulators.py
```diff
@@ -641,6 +663,160 @@ def ost2raven(self, ops):
...
+
+    identifier = "sacsma"
+    templates = tuple((Path(__file__).parent / "raven-sacsma").glob("*.rv?"))
+
+    params = namedtuple(
+        "SACSMAParams", ", ".join(["par_x{:02}".format(i) for i in range(1, 22)])
+    )
+
+    @dataclass
+    class HRU(HRU):
```
cjauvin, line 679: It doesn't make sense to have a derived class named the same as its parent. Please use a more specialized name for the child class.

ravenpy/models/emulators.py
```diff
@@ -641,6 +663,160 @@ def ost2raven(self, ops):
...
+            params=SACSMA.params(*((None,) * len(SACSMA.params._fields))),
+            land_use_classes=(
+                LU("FOREST", impermeable_frac=0.0, forest_coverage=0.02345),
+            ),
+        )
+        self.rvh = RVH(hrus=(SACSMA.HRU(),))
+        self.rvt = RVT(**{k: nc() for k in std_vars})
+        self.rvi = RVI(
+            evaporation="PET_OUDIN", rain_snow_fraction="RAINSNOW_DATA"
+        )  # would like to use RAINSNOW_HBV??
```
cjauvin, line 698: Try to put the comments on top of lines instead of at the end, as Black's formatting makes long line a bit weird.

cjauvin (approved):

ravenpy/models/emulators.py
huard: To be clear, my question is "why do you need copies of parameters with different names?" Could you instead use the original name in the template ? I'm ok with the code as-is if it makes your life easier, it's just a question.

ravenpy/models/emulators.py
```diff
@@ -440,11 +455,19 @@ def derived_parameters(self):
...
     identifier = "blended"
     templates = tuple((Path(__file__).parent / "raven-blended").glob("*.rv?"))
 
     params = namedtuple(
-        "BLENDEDParams", ", ".join(["par_x{:02}".format(i) for i in range(1, 36)]+["par_r{:02}".format(i) for i in range(1, 9)])
+        "BLENDEDParams",
+        ", ".join(
+            ["par_x{:02}".format(i) for i in range(1, 36)]
+            + ["par_r{:02}".format(i) for i in range(1, 9)]
+        ),
```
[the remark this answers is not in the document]
huard: These are model parameters. How we name them is a convention specific to each model, so it makes sense to define those inside the Emulator class.

ravenpy/models/emulators.py
huard: That's on me, but I'm not sure I agree. The scope is different, so it's another object. We could have defined the original HRU class inside a base Emulator class, and we'd just be subclassing this attribute.

ravenpy/models/emulators.py
julemai: In the Ostrich setup they will be named `pow_x01`. So, I like that at least one can see that this the `10**x1` and not actually `x1`. The ranges of parameters we provide is for `x1` while Raven needs `10**x1` as a parameter. I would find it very confusing if it would be called `x1` in the model setup. I am totally fine with how they look like right now. In a very early version of the predecessor of RavenPy we actually passed `10**x1` there and the user had to provide `x1`. At one point we decided that this is very confusing for the users and switched it to the values that are actually used by Raven (i.e. `pow_x01=10**x1`). 

It's indeed very easy for me as it is.

ravenpy/models/emulators.py
julemai: Sure but then we should change this in ALL other tests too.

ravenpy/models/emulators.py
julemai: I let you two decide as I honestly don't really know what this is doing anyway. :)

ravenpy/models/emulators.py
julemai: Sorry. Should remove this comment entirely. It's a remainder of my testing

ravenpy/models/emulators.py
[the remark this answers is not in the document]
julemai: Yes, please don't change those names. I use them throughout the entire documentation and in every template.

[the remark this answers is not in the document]
julemai: Please no magic. For once. :)

ravenpy/models/emulators.py
huard: But shouldn't it be `self.rvd["POW_X01"] = np.log10(self.rvp.params.par_x01)`

julemai: No. Because the user provides the value `10**x1` and not `x1`. The problem is that this is a parameter that spans multiple orders of magnitudes (like 0.00001 to 10,000). So, for sampling of this parameter in calibration it is much better to sample in the logarithmic space (like -5 to 4). The first one (like 0.00001 to 10,000) is the one Raven needs (here `pow_x1`) and is hence provided when we run Raven. BUT during calibration and in the Ostrich setup this parameter is the logarithmic one (like -5 to 4) and hence `x1`.

huard: Now I get it! Sorry it took so long.

ravenpy/models/emulators.py
julemai: Done.

ravenpy/models/emulators.py
julemai: No need to be sorry. It is really convoluted because we have Raven itself and then also the Ostrich setup.

cjauvin:

PR #79, opened 2021-03-15
Add HYPR model

julemai:
The HYdrological model for Prairie Region (HYPR) is based on the conceptual Hydrologiska Byrans Vattenbalansavdelning (HBV)-light model. HBV is modified to work in the prairies by incorporating a conceptual lateral-flow component to represent the pothole storage complexities (PDMROF). HYPR can be used for prairie streamflow simulation.

All tests pass.


ravenpy/models/emulators.py
```diff
@@ -641,6 +659,185 @@ def ost2raven(self, ops):
...
+    conceptual lateral-flow component to represent the pothole storage
+    complexities. HYPR can be used for prairie streamflow simulation.
+
+    References
+    ----------
+    Mohamed I. Ahmed, Amin Elshorbagy, and Alain Pietroniro (2020).
+    Toward Simple Modeling Practices in the Complex Canadian Prairie
+    Watersheds. Journal of Hydrologic Engineering, 25(6), 04020024.
+    doi: 10.1061/(ASCE)HE.1943-5584.0001922.
+    """
```
huard, line 676: Nice !

ravenpy/models/raven-hypr/raven-hypr.rvc
```diff
@@ -0,0 +1,13 @@
...
+:CreationDate      Feb 2021
+#
+# Emulation of the HYPR model for Salmon River near Prince George
+#------------------------------------------------------------------------
+
+#  Raven Initial Conditions file
+
+:UniformInitialConditions SOIL[0] 10
+:UniformInitialConditions SOIL[1] 10
+:UniformInitialConditions DEPRESSION 20
```
huard, line 13: We'll probably want to eventually template the initial conditions to be able to control them from RavenPy, ie using  
```
{hru_states_cmd}

{basin_states_cmd}
```
But we can do that later.

huard (approved):

ravenpy/models/raven-hypr/raven-hypr.rvc
julemai: Probably. We just need to make sure that we enable that these can be calibratable parameters. On top of being calibratable, they are usually derived parameters (like half soil depth in mm) in the calibration setup. But I am sure we can make this work. 

I just tried to use something similar in the CANADIANSHIELD model with the `area`. It us within a `hru_cmd` and it is a derived parameter... So, it's possible; just too complex for me to find the correct syntax and way to use it best.

ravenpy/models/emulators.py
```diff
@@ -641,6 +659,185 @@ def ost2raven(self, ops):
...
+
+    identifier = "hypr"
+    templates = tuple((Path(__file__).parent / "raven-hypr").glob("*.rv?"))
+
+    params = namedtuple(
+        "HYPRParams", ", ".join(["par_x{:02}".format(i) for i in range(1, 22)])
+    )
+
+    @dataclass
+    class HRU(HRU):
```
cjauvin, line 686: The derived class should have a more specialized name.

cjauvin (approved):

ravenpy/models/emulators.py
huard: Like HYPRHRU ? 
From a user perspective, it would show up like:
`HYPR.HYPRHRU` which I tought was a bit redundant, but willing to be convinced this is clearer than `HYPR.HRU`.

cjauvin: I just thought the chain of inheritance becomes hard to follow at this point: an internal class that derives (and has the same name as) from a typedef of the internal class of another class in another module..

huard: Another option would be to redefine it from scratch, but I'm not sure this would be an improvement. 

The thing that bothers me most with the current setup is that some fields are still mutable. 
```
        land_use_class: str = "OPEN_1"
        veg_class: str = "FOREST"
        soil_profile: str = "DEFAULT_P"
        aquifer_profile: str = "[NONE]"
        terrain_class: str = "[NONE]"
```
are just default values, but it still possible to modify them. I would like to be able to say: these are immutable class properties.

huard: This kind of idea is discussed here: https://stackoverflow.com/questions/58532383/dataclass-style-object-with-mutable-and-immutable-properties
but I'm not in love with the solution proposed.

julemai: I am fine with whatever you decide here. We definitely want these "default" values to be changeable by the users as, for example, the routing product provides different names for this in any version I am getting.

huard: You mean that within the HYPR, model these names could change ? In other words, these are not set by the emulator itself, but by the product used to generate the HRU/subbasin config ? In that case my comment is moot. Darn, I thought I had found a solution... 

For reference, pydantic is a library doing input parsing to conform to a schema, and can automatically detect to which schema an object matches. It also allows some class properties to be set as immutable. https://pydantic-docs.helpmanual.io/usage/schema/#field-customisation

julemai: Technically they can change. I am pretty sure that the `soil_profile` name can change. It just needs to be consistent with which name is used in the other RV* files. I am not sure that we template those names there yet. So, it is good that they are fixed for now. But the names do no have any magic power. It is (for almost all of them) just a label.

huard: Ok. At the moment, I think some of these fields are hard-coded in some places, but soft-coded in others. So my intent was to make sure they were hard-coded everywhere to avoid inconsistencies. Now if in the routing case these labels might vary, we'll need "soft-coded AND consistent." But I'm happy to defer this issue, thanks for the clarifications.

julemai: You are absolutely right about soft-/hard-coding of some fields. We just never had this in mind when we started here. And I didn't know myself that these labels could change. At one point we will need to move to "soft-coded AND consistent". Probably not right now.

cjauvin:

PR #80, opened 2021-03-17
Adding model for Canadian Shield

julemai:
All tests pass.

Two TODO's because I couldn't figure out how to do that:
* use of landuse `lu` passed by test in `class CANADIANSHIELD(Raven)`
* use of `area` in `class CANADIANSHIELD_OST(Ostrich, CANADIANSHIELD)`

The Canadian Shield model is a useful configuration of Raven for Canadian shield basins characterized by shallow soils atop rock, with ample exposed rock and lakes.


julemai:
Hi @huard! This is the last of the four models we wanted to add to RavenPy in this round. The difficulty here is that the one sub-basin has 2 HRUs (ORGANIC and BEDROCK). @cjauvin helped me to get this working. 

There is two things that we might want to get done. I listed them as TODO's above and will mark the part in the code with a comment. I think both are pretty straight forward. I just couldn't figure out which syntax to use to obtain the area and the land-use class parameters set in the test (by the user).

ravenpy/models/ostrich-canadianshield/ostIn.txt
```diff
@@ -0,0 +1,114 @@
...
+  par_pow_x08      1 par_x08         exp    10.0 1.00 1.00 0.00 free
+  par_pow_x09      1 par_x09         exp    10.0 1.00 1.00 0.00 free
+  par_half_x01     1 par_x01         linear 500.0     0.00      free
+  par_half_x02     1 par_x02         linear 500.0     0.00      free
+  par_half_x03     1 par_x03         linear 500.0     0.00      free
+  #
+  # AREA HRU1=SOILP_ORG (km2)
+  # soil organic area = x34 * AREA
+  # area SB1 = {area}
+  par_area_organic_SB1 1 par_x34         linear  {area}      0.0    free
```
julemai, line 74: This is the reason why the Ostrich setup needs the area (here: 4250.6 km2) as a parameter...

ravenpy/models/emulators.py
```diff
@@ -641,6 +659,226 @@ def ost2raven(self, ops):
...
+    templates = tuple(_p.glob("model/*.rv?")) + tuple(_p.glob("*.t??"))
+
+    def __init__(self, *args, **kwds):
+        super().__init__(*args, **kwds)
+        self.rvi.suppress_output = True
+        self.txt = Ost(
+            algorithm="DDS",
+            max_iterations=10,
+            # TODO: This needs to receive the area (hru_default_values["area"])
+            area=4250.6,
```
julemai, line 780: The `area` should not be hard-coded here

ravenpy/models/emulators.py
```diff
@@ -641,6 +659,226 @@ def ost2raven(self, ops):
...
+        super().__init__(*args, **kwds)
+        self.rvp = RVP(
+            params=CANADIANSHIELD.params(
+                *((None,) * len(CANADIANSHIELD.params._fields))
+            ),
+            # TODO:: use here LU set by user in test_candianshield
+            land_use_classes=(
+                LU("FOREST", impermeable_frac=0.0, forest_coverage=0.02345),
+            ),
+        )
```
julemai, line 707: The `LU` should not be hard-coded here.

huard: See this rather as the default value, not something that is hard-coded. In principle, the values you pass to the model at execution should override this.

ravenpy/models/ostrich-canadianshield/ostIn.txt
huard: Could we create this in the `derived_parameters` method ?
I see that you've created slots for `area_frac_organic` and `area_frac_bedrock`. 
My understanding is that we could write something like: 

```
for hru in self.rvh.hrus:
    if type(hru) is CANADIANSHIELD.HRU_ORGANIC:
        self.rvd['area_frac_organic'] = hru.area      
    elif type(hru) is CANADIANSHIELD.HRU_BEDROCK:
        self.rvd['area_frac_bedrock'] = hru.area
    else:
        raise ValueError("HRU instance not recognized.")
```

ravenpy/models/emulators.py
huard: If you're saying that the LU given in the test does not override the default value given in the CANADIANSHIELD class, then this is a bug for me to fix.

ravenpy/models/ostrich-canadianshield/ostIn.txt
julemai: @huard Maybe? But the areas are not yet known. The area is the total area (4250.6) multiplied with par_x34 for ORGANIC and then 4250.6*(1-par_34) for BEDROCK. The current implementation works in both setups (raven and ostrich) thanks to @cjauvin for working this out together with me. 

I just wouldn't know where `hru.area` is coming from and how to connect then this one with `par_x34`... You know what I mean?

ravenpy/models/emulators.py
julemai: I haven't actually tested if it would be overwritten when I initiate it with a different value. Let me check.

julemai: Of course it's not a bug :) 

It works as you said. This is overwritten by whatever I pass to the model at execution. I think my logic was just exactly the other way around. So, it's all good.

ravenpy/models/ostrich-canadianshield/ostIn.txt
cjauvin: Just a note: I think it might be better if we use `isinstance` instead of `type is` because the former works with inheritance, which we use a lot in RavenPy.

cjauvin (approved):

ravenpy/models/ostrich-canadianshield/ostIn.txt
huard: You're right.

cjauvin:

PR #84, opened 2021-04-01
Climpred notebook for ESP forecast

huard:
Notebook could take some love and more interesting graphics, but it's a start. 

Branched off add-routing-to-emulator, so let's wait for it to be merged.


richardarsenault:
I'm getting an error while running the notebook. The fact that there is no rvc seems to be making it fail. When I add one, it works. Do you have the same happening on your end?

huard:
@richardarsenault Should be fixed now. Sorry, stupid mistake.

environment.yml
```diff
@@ -21,7 +21,8 @@ dependencies:
   - owslib
   - pandas
   - pip
-  - proj<8.0.0
+  - proj
```
Zeitsperre: Good call. This was causing problems a few months ago but was dealt with recently.

PR #87, opened 2021-04-07
Enable GitHub CI

### LINKED ISSUE #67: Enable CI with Github actions

Zeitsperre:
Closes #67

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ deps =
 ; requirements.txt with the pinned versions and uncomment the following line:
 ;     -r{toxinidir}/requirements.txt
 commands =
-    python -m pip install --no-user {opts} . --install-option="--with-binaries"
+    python -m pip install --no-user --verbose . --install-option="--with-binaries"
     git clone https://github.com/Ouranosinc/raven-testdata {envtmpdir}/raven-testdata
     env RAVENPY_TESTDATA_PATH={envtmpdir}/raven-testdata pytest --cov ravenpy
     - coveralls
--- a/tox.ini
+++ b/tox.ini
@@ -25,6 +25,8 @@ allowlist_externals = make
 setenv =
     PYTEST_ADDOPTS = "--color=yes"
     PYTHONPATH = {toxinidir}
+passenv =
+    GDAL_VERSION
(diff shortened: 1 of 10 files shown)


cjauvin (approved):

Zeitsperre:
Still failing. I need to somehow force the raven/ostrich binaries to install properly. We're nearly there.

tox.ini
```diff
@@ -27,11 +27,18 @@ setenv =
     PYTHONPATH = {toxinidir}
 extras = dev
 download = true
-install_command = python -m pip install --no-user {opts} {packages}
+install_command =
+    python -m pip install --no-user {opts} {packages}
+    python -m pip install --no-user {opts} {packages} --install-option="--with-binaries"
```
huard, line 32: Why twice the same command ? 
Don't we want to run tests inside the conda environment ?

huard (approved):

tox.ini
Zeitsperre: The reason for calling the command twice is due to the fact that calling `--install-option` during a pip installation disables `wheels`. For most of the dependencies we have, `wheels` are preferable, so installing all those are done in the first call, and compiling raven/ostrich is performed in the second.

Zeitsperre:
@huard The CI is configured for `pip` now. Any failures from the builds are due to bugs/errors that would be present in a clean system. Should I add a build for `conda` while I'm at it?

huard:
The Raven server environment is installed with conda, so I suggest we do that. However, I suggest we open a new issue and PR for this, as it is less of a priority now that this is done.

tox.ini
huard: But I believe the second command won't install ravenpy since it's already installed. You need to uninstall ravenpy before installing it with the binaries.

huard:
If the CI is activated, I suggest we merge it and fix the test failures in the next PR.

huard:
Also, I'd be comfortable reducing the requirements to a single python version (e.g. 3.8) for now, to be extended later as ravenpy matures.

huard:
there is a typo in the conda title "fro Python3.8"

PR #90, opened 2021-04-08
Release 0.4.0 (routing update)

cjauvin:
Please update the HISTORY file.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,9 @@ History
 
 * GR4JCN emulator now supports routing mode.
 * Add BLENDED emulator.
+* Added support for `tox`-based localized installation and testing with python-pip.
+* Now supporting Python 3.7, 3.8, and 3.9.
+* Build testing for `pip` and `conda`-based builds with GitHub CI.
 
 0.3.1
 -----
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 History
 =======
 
+0.4.0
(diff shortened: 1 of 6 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+0.4.0
+-----
+
+* GR4JCN emulator now supports routing mode.
+* Add BLENDED emulator.
+
```
Zeitsperre, line 10:
```suggestion
* Added support for `tox`-based localized installation and testing with python-pip.
* Now supporting Python 3.7, 3.8, and 3.9.
* Build testing for `pip` and `conda`-based builds with GitHub CI.

```

HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+0.4.0
+-----
+
```
huard, line 7:
```suggestion
This is an interim version making one step toward semi-distributed modeling support. Model configuration is still in flux and will be significantly modified with 0.5. The major change in this version is that model configuration requires the creation of HRU objects, instead of simply passing area, latitude, longitude and elevation. 
```

huard (approved):

HISTORY.rst
cjauvin: Actually we still have an adapter to support the old HRU quatuor.. :violin: :musical_keyboard:

huard: adjusted

PR #98, opened 2021-04-14
Add RavenError and explicit Raven_errors.txt parsing

cjauvin:
This tries to make the RavenC error processing mechanism more explicit by doing two things:

1. Parse all the `Raven_errors.txt` messages
2. Coalesce the messages of type `ERROR` into a `RavenError` custom exception

The end goal of this is to catch this exception in RavenWPS and return it to the user as a `pywps.app.exceptions.ProcessError`, for friendlier debugging.


huard (approved):
Excellent !

PR #102, opened 2021-04-26
update to newest black (v21.4b0) and use python3.7 import conventions

Zeitsperre:
This PR does two things:
- Updates the version of black in the pre-commit configuration. For information on the new features enabled by default, see: https://github.com/psf/black/releases/tag/21.4b0
- Updates the `isort` and `black` target python to the new lowest supported version (python3.7).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
 
 repos:
 -   repo: https://github.com/asottile/pyupgrade
-    rev: v2.10.0
+    rev: v2.13.0
     hooks:
     -   id: pyupgrade
         language_version: python3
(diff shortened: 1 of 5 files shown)


Zeitsperre:
Interesting: https://github.com/CSHS-CWRA/RavenPy/runs/2440566540#step:6:382

Looking into this now.

Zeitsperre:
This error appears to be related to the environment that numpy was compiled in: https://stackoverflow.com/questions/29206612/difference-between-data-type-datetime64ns-and-m8ns. 

Would casting this to a specific datetime format within the test help, or is there a more dangerous bug lurking in the shadows?

.pre-commit-config.yaml
```diff
@@ -21,25 +21,25 @@ repos:
...
         language_version: python3
         args: ['--config=setup.cfg']
 -   repo: https://github.com/timothycrosley/isort
-    rev: 5.7.0
+    rev: 5.8.0
     hooks:
     -   id: isort
         language_version: python3
-        args: ['--profile', 'black', '--filter-files', '--py', '36']
+        args: ['--profile', 'black', '--filter-files', '--py', '37']
```
cjauvin, line 40: I would suggest we get rid of the `--py` option, I don't see the point of keeping it.

cjauvin (approved):

cjauvin:
Does this need to be accompanied by a run of the updated Black on the entire repo? And is there something that we must do locally to update our pre-commit setup?

.pre-commit-config.yaml
Zeitsperre: Okay. The only change dropping it would do is we _might_ see some changes happen without our knowledge if we use some new library in the standard lib that's been backported to an older python. This is a really, really specific case, so that's fine.

Zeitsperre:
Nothing at all needs to be done locally once the change is in the pre-commit configuration. The run over all files has already been done (only one file affected, so good job, us). Set it and forget it.

PR #106, opened 2021-05-18
Add Julie's new emulators (post refactoring)

cjauvin:
This corresponds and should be equivalent to this pre-refactoring PR: https://github.com/CSHS-CWRA/RavenPy/pull/77

Even though there are tests, the ideal would probably be to inspect the generated config files to make sure they are fully valid (the way to do this is as before: pass a workdir path to the emulator constructor in the tests).

Two similar PRs for the other models should follow soon.


julemai:
Got it. But probably won't have time to look at that in detail before end of the week.

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,704 @@
...
+        # R V P #
+        #########
+
+        rvp_tmpl = """
+        #########################################################################
+        :FileType          rvp ASCII Raven 3.0.4
+        :WrittenBy         James Craig & Juliane Mai
+        :CreationDate      Feb 2020
+        #
+        # Emulation of SAC-SMA simulation of Salmon River near Prince George
```
huard, line 86:
```suggestion

```

huard (approved):
Looks clean and good !

ravenpy/models/emulators/sacsma.py
cjauvin: thanks, done.

cjauvin:
@huard @julemai Finally I will add all the new emulators in this PR, contrary to what I have said. So I just added the HYPR model, please note:
1. Following our discussion I have added a `DEPRESSION` state variable to the `HRUState` struct (the default value is 0mm).
2. For the RVH I have implemented the `:SubBasinProperties` as an additional template section (along with an "extra attribute" for the value): this is debatable, as it could also be implemented as a proper `ravenpy.config.Command` dataclass

cjauvin:
@julemai @huard The 3 new models have been added, this PR is thus ready for final review.

ravenpy/models/emulators/canadianshield.py
```diff
@@ -0,0 +1,616 @@
...
+               :Units,               -,         -,               -,             -,                   -,                -,          -,               -,               -,
+              TOPSOIL,             1.0,   par_x07,             0.0,           0.0,             par_x12,  par_sum_x05_x06,    par_x05,         par_x13,         par_x04,
+               VADOSE,             1.0,       0.0,     par_pow_x08,       par_x10,                 0.0,              0.0,        0.0,         par_x14,             0.0,
+          FRACBEDROCK,             1.0,       0.0,     par_pow_x09,       par_x11,                 0.0,              0.0,        0.0,             0.0,             0.0,
+        #     TOPSOIL,             1.0,  para_x07,             0.0,           0.0,            para_x12, para_sum_x05_x06,   para_x05,        para_x13,        para_x04,
+        #      VADOSE,             1.0,       0.0,    para_pow_x08,      para_x10,                 0.0,              0.0,        0.0,        para_x14,             0.0,
+        # FRACBEDROCK,             1.0,       0.0,    para_pow_x09,      para_x11,                 0.0,              0.0,        0.0,             0.0,             0.0,
```
huard, line 389: Do we need to keep those commented lines ?

ravenpy/models/emulators/canadianshield.py
```diff
@@ -0,0 +1,616 @@
...
+        self.config.rvp.derived_params = CANADIANSHIELD.DerivedParams(
+            SUM_X05_X06=params.par_x06,
+            SUM_X16_X17=params.par_x17,
+            POW_X08=params.par_x08,
+            POW_X09=params.par_x09,
+        )
+
+        # We must have two HRUs with equal areas
+        assert len(self.config.rvh.hrus) == 2
+        assert self.config.rvh.hrus[0].area == self.config.rvh.hrus[1].area
```
huard, line 281: Replace by try/expect statements. assert statements are removed with the -O optimization flag.

ravenpy/models/emulators/canadianshield.py
```diff
@@ -0,0 +1,616 @@
...
+
+        self.config.rvh.hrus[0].area *= params.par_x34
+        self.config.rvh.hrus[1].area *= 1.0 - params.par_x34
```
huard, line 284: Is there a risk derived_parameters gets called multiple times and the areas won't make sense ? 
How will this work in distributed mode ? 
Should we spawn the twin hru automatically instead ?

huard: I guess more broadly, maybe derived_parameters should not affect the internal state, and instead return a modified copy of the internal state?

ravenpy/models/emulators/canadianshield.py
```diff
@@ -0,0 +1,616 @@
...
+        soil_profile: str = "SOILP_BEDROCK"
+        aquifer_profile: str = "[NONE]"
+        terrain_class: str = "[NONE]"
+
+    def __init__(self, *args, **kwds):
+        super().__init__(*args, **kwds)
+
+        self.config.update(
+            identifier="canadianshield",
+            # hrus=(HYPR.HRU(),),
```
huard, line 94: No HRU is initialized ?

huard: We're using the default one ?

huard (approved):
There are a couple of commented lines in the templates that maybe could be removed, but otherwise LGTM.

tests/test_emulators.py
```diff
@@ -438,7 +438,7 @@ def test_run(self):
         np.testing.assert_almost_equal(d["DIAG_NASH_SUTCLIFFE"], -0.117301, 4)
 
     def test_evaluation(self):
-        model = GR4JCN("/tmp/testeval")
+        model = GR4JCN()
```
huard, line 441: Thanks !

ravenpy/models/emulators/canadianshield.py
julemai: I would leave them in. Otherwise, it is really hard for the user to identify which parameter is which once the values get filled in. The commented lines won't get filled in because they contain, for example, `para_x07` and not `par_x07`.

huard: You're referring to the "filling-in" by Ostrich ?

ravenpy/models/emulators/canadianshield.py
julemai: In this example the sub basin will be split into two HRUs (bedrock and not-bedrock). Hence, the areas will always need to sum up to the actual area. This split will only be performed for land-HRUs and never for lake-HRUs. So, we are safe also in our distributed setups.

ravenpy/models/emulators/canadianshield.py
julemai: No, also in general. 

If the user downloads the RV* files with the parameters set he/she might have chosen. There would be no way for the user to know where parameter x7 is located in those files when you remove those comment lines.

huard: I see your point. Makes sense.

ravenpy/models/emulators/canadianshield.py
cjauvin: Yes that is an "oubli", can you confirm that the fix is ok please (not 100% sure).

ravenpy/models/emulators/canadianshield.py
```diff
@@ -0,0 +1,619 @@
...
+        # SOILP_BEDROCK, 3, TOPSOIL, para_x01, VADOSE, 0.00,     FRACBEDROCK, para_x03,
+        :EndSoilProfiles
+
+        #-----------------------------------------------------------------
+        # Land Use Classes
+        #-----------------------------------------------------------------
+        # :LandUseClasses,
+        #   :Attributes,        IMPERM,    FOREST_COV,
+        #        :Units,          frac,          frac,
+        #   FOREST,                0.0,       0.02345 ## forest_frac,
```
julemai, line 356: Maybe change to:
```
# FOREST, 0.0, forest_frac,
```

ravenpy/models/emulators/hypr.py
```diff
@@ -0,0 +1,611 @@
...
+        #########
+        # R V P #
+        #########
+
+        rvp_tmpl = """
+        # tied parameters:
+        # (it is important for OSTRICH to find every parameter place holder somewhere in this file)
+        # (without this "para_x05" and "para_x06" wouldn't be detectable)
+        #    para_pow_x5     = {derived_params.POW_X05}     =  10^(para_x05)       = 10^{params.par_x05}
+        #    para_pow_x6     = {derived_params.POW_X06}     =  10^(para_x06)       = 10^{params.par_x06}
```
julemai, line 100: This does not work. Both values (`derived_params.POW_X05/06` and `params.par_x05/06`) get filled in with same value...

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,672 @@
...
+        #                                            model); mm/d/C; default=1.5
+        # para_x18 0.3    20.0     # GAMMA_SHAPE;    used to build unit hydrograph, LandUseParameterList; none ; default=1.0
+        # para_x19 0.01   5.0      # GAMMA_SCALE;    used to build unit hydrograph, LandUseParameterList; none ; default=1.0
+        # para_x20 0.8    1.2      # RAINCORRECTION; Muliplier to correct rain, Gauge; none ; default=1.0
+        # para_x21 0.8    1.2      # SNOWCORRECTION; Muliplier to correct snow, Gauge; none ; default=1.0
+
+        # tied parameters:
+        # (it is important for OSTRICH to find every parameter place holder somewhere in this file)
+        # (without this parameters that are used to derive parameters wouldn't be detectable by Ostrich)
+        #    para_x04_mm       = par_soil0_mm     = para_x04 * 1000       =  par_x04 * 1000
```
julemai, line 136: should be:
```
para_x04_mm = par_soil0_mm = para_x04 * 1000 = {params.par_x04} * 1000
```
such that values get filled in

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,672 @@
...
+        #    para_x06_mm       = par_soil0_mm     = para_x06 * 1000       =  par_x06 * 1000
```
julemai, line 137: should be:
```
para_x06_mm = {derived_params.par_soil2_mm} = para_x06 * 1000 = {params.par_x06} * 1000
```
such that values get filled in (see also typo for `par_soil0_mm` instead of `par_soil2_mm`)

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,672 @@
...
+        #    para_bf_loss_frac = PAR_BF_LOSS_FRAC = para_x12/(1+para_x12) = par_x12/(1+par_x12)
+        #    para_pow_x01      = POW_X01          = 10^(para_x01)         = 10^par_x01
+        #    para_pow_x02      = POW_X02          = 10^(para_x02)         = 10^par_x02
+        #    para_pow_x03      = POW_X03          = 10^(para_x03)         = 10^par_x03
+
```
julemai, line 142: above 4 parameters get also not filled in properly. Two out of the four values need to be templated using something like `{}`

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,672 @@
...
+        params = cast(SACSMA.Params, self.config.rvp.params)
+        self.config.rvp.derived_params = SACSMA.DerivedParams(
+            PAR_BF_LOSS_FRAC=params.par_x12,
+            POW_X01=params.par_x01,
+            POW_X02=params.par_x02,
+            POW_X03=params.par_x03,
+        )
+
+        soil0 = params.par_x04 * 1000.0
```
julemai, line 286: Maybe it would be helpful to name them here as we call them above (i.e. `par_soil0_mm`)? Same for the one in next line.

ravenpy/models/emulators/sacsma.py
```diff
@@ -0,0 +1,672 @@
...
+        # tied parameters:
+        # (it is important for OSTRICH to find every parameter place holder somewhere in this file)
+        # (without this parameters that are used to derive parameters wouldn't be detectable by Ostrich)
+        #    para_x04_mm       = par_soil0_mm     = para_x04 * 1000       =  par_x04 * 1000
+        #    para_x06_mm       = par_soil2_mm     = para_x06 * 1000       =  par_x06 * 1000
+        #    para_bf_loss_frac = par_bf_loss_frac = para_x12/(1+para_x12) = par_x12/(1+par_x12)
+        #    para_pow__x01     = pow_x01          = 10^(para_x01)         = 10^par_x01
+        #    para_pow__x02     = pow_x02          = 10^(para_x02)         = 10^par_x02
+        #    para_pow__x03     = pow_x03          = 10^(para_x03)         = 10^par_x03
+
```
julemai, line 382: See my comments above. There are some of them that need to be templated such that sampled and derived values get filled in.

julemai (changes requested):
There are a few minor things that would improve the user experience when downloading the setups.

The only major comment I have is that the header for all rv* files got lost in this process. The header contained who developed those models in the Raven framework and I think we should give credit to those people by including again a proper header (potentially also containing the Raven version this was build with). What do you think?

ravenpy/models/emulators/hypr.py
cjauvin: I don't understand what you mean, those are two seperate variables, and it's a comment. You mean the comment is not generated properly?

julemai: That's correct; it is a comment. But it is setup like you can see the sampled and the derived parameters such that one can easier check what actually happened and it is really convenient for debugging and checking. HERE: the derived parameter is POW_X05 (which is 10**x05).

So, it would be great to see (let's say x05 is sampled to be 2):
`para_pow_x5     = 100.0    =  10^(para_x05)       = 10^2.0`

You see what I mean? It worked in the examples I provided. This might happen at more places. This is one where I spotted it

cjauvin: I went back in your own code. In your test you have this:

https://github.com/CSHS-CWRA/RavenPy/blob/add-hypr-model/tests/test_hypr.py#L30-L31

so it seems (as you say in the comments) that `par_x05` and `par_x06` are actually 10 to their power, from the start.. which would make sense because later you have this:

https://github.com/CSHS-CWRA/RavenPy/blob/add-hypr-model/ravenpy/models/emulators.py#L718-L720

(I always wondered what was the point of having the part on the right commented out, because without it `par_x05` is just trivially equal to `POW_X05`). I asked @huard but he could not explain it to me. I have observed that you have the same kind of logic in a few other places.

So the result is that `par_x05` == `POW_X05` (idem for `06`), from start to finish, and so when RavenPy renders the template values that you have in your RVP, namely:

https://github.com/CSHS-CWRA/RavenPy/blob/add-hypr-model/ravenpy/models/raven-hypr/raven-hypr.rvp#L13-L14

the resulting RVP is, when I run it (which I just did, using the code in your PR):

```
#    para_pow_x5     = 0.4650976     =  10^(para_x05)       = 10^0.4650976
#    para_pow_x6     = 0.1177004     =  10^(para_x06)       = 10^0.1177004
```

which does not make sense to me. I realize it's probably me who is confused, but please help me understand what I'm getting wrong.

cjauvin: Just to clarify: these are RV comments so they do not affect the actual results in any way of course, but still the comments do not make sense (arithmetically speaking).

julemai: `par_x05` is NOT equal to `POW_X05`. That is VERY important. When we just run Raven, we pass a set of parameters- original one AND derived ones. When we use OSTRICH, all parameters are the original ones and Ostrich is doing the derivation of the parameters. Hence all the comments in my codes to make clear if something is derived or not. The piece above is for the forward run of Raven (not Ostrich) and hence you are right, `par_x05` is not actually known. So leave it as it is. But please don't rename POW_x05 to par_x05 because that is not correct.

julemai: This all won't change the results since this is all comments. But you can see how confusing this already is to you, so it will be even more confusing to the users if there is written:
`0.4650976 = 10^(para_x05) = 10^0.4650976`. We need to remove the last piece then and make it only `0.4650976 = 10^(para_x05)`. Makes sense?

cjauvin:
@julemai @huard My last commit represents the state of my understanding and thinking following our very productive early evening discussion. So my idea is that since it's only in the Ostrich world that there's an ambiguity between `par_x05` and `pow_x05` (for the sampling efficiency reason you explained), I decided to remove entirely any notion of "derived params" and "pow_whatever" from the `HYPR` (i.e. pure Raven) configuration. In the Raven `HYPR` world, `par_x05` means only one thing, and there's no ambiguity whatsoever. In the `HYPR_OST` world, where there is such an ambiguity, I explain it clearly in the `ost2raven` method, which I also refactored so that it makes better use of the new tools I introduced in my refactoring. What do you think?

julemai:
I am honestly not so sure about this.

1. We would need to roll out this concept for all other 7 models now too.
2. I think it is very confusing that `par_x05` will have different meanings depending if it is Ostrich or if it is Raven.
3. Sometimes both variables- the original AND the derived ones are needed in a setup. So, this concept wouldn't work there.

I think we should just leave it as it was and either removing those comments in the RV* files where we can't fill in the proper values because they are not available or deriving `par_x05` from the provided `pow_x05`. Either is fine for me but I would like these parameters be called the same in Raven and in the Ostrich setup.

cjauvin:
@julemai I must admit, I simply do not understand what you want. I thought I understood yesterday while we were speaking, but when I went into the code the only thing that made sense to me was I did afterward (last commit). So can I kindly ask you to simply provide the modifications you want, and push them to this PR? I really think this will be the simplest strategy. If you cannot or don't have the time (totally understandable) I will need to merge as is, because we need to move on eventually.

ravenpy/models/emulators/hypr.py
```diff
@@ -0,0 +1,602 @@
...
+        # R V P (OST TMPL) #
+        ####################
+
+        rvp_tmpl = """
+        # tied parameters:
+        # (it is important for OSTRICH to find every parameter place holder somewhere in this file)
+        # (without this "para_x05" and "para_x06" wouldn't be detectable)
+        #    para_pow_x5     = pow_x05     =  10^(para_x05)       = 10^par_x05
+        #    para_pow_x6     = pow_x06     =  10^(para_x06)       = 10^par_x06
+
```
julemai, line 334: @cjauvin I was concerned that the recent adjustments you made would lead to a correct population of those commented lines. Hence, my suggestion was to modify the previous version of `derived_parameters`:

```
    def derived_parameters(self):
        params = cast(HYPR.Params, self.config.rvp.params)
        self.config.rvp.derived_params = HYPR.DerivedParams(
            POW_X05=params.par_x05,
            POW_X06=params.par_x06,
        )
```
to the following version (not sure if the syntax would work...):
```
    def derived_parameters(self):
        params = cast(HYPR.Params, self.config.rvp.params)
        self.config.rvp.derived_params = HYPR.DerivedParams(
            POW_X05=params.par_x05,
            POW_X06=params.par_x06,
            par_x05 = np.log10(params.par_x05),
            par_x06 = np.log10(params.par_x06),
        )
```

This way all four parameters would be known and contain the correct values. This is important since sometimes Ostrich will need both versions of the value (e.g., POW_x05 and par_x05). Not in this example but for other models...

I am just afraid that the current implementation would not be generalizable for the other 7 models.

cjauvin: Hi @julemai! What you propose would certainly work for sure, but I have 3 questions:

1. Why should we be concerned about Ostrich-related stuff in the emulator (I'm talking about the specific `hypr.HYPR` class, which is Raven-only, as opposed to `hypr.HYPR_OST`, which is Raven+Ostrich) that is (to my knowledge) completely devoid of any Ostrich stuff? (this was the spirit of my last commit, to be clear)
2. Why do you say that this approach wouldn't generalize to the other models?
3. .. I forgot my third point..

PR #113, opened 2021-05-26
Added support for EvaluationPeriod

huard:
Let's users compute diagnostics on user-defined periods. 

API change: model.diagnostics will return a dict of list instead of a dict of scalars. 

TODO: Tie into Ostrich.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,7 @@ History
 0.6.0
 -----
 
-* Added support for EvaluationPeriod commands. Note that as a result of this, the model diagnostic property contains lists instead of a scalars.
+* Added support for EvaluationPeriod commands. Note that as a result of this, the model's `diagnostics` property contains one list per key, instead of a single scalar. Also note that for calibration,  Ostrich will use the first period and the first evaluation metric.
 
 0.5.2
 -----
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+0.6.0
+-----
+
+* Added support for EvaluationPeriod commands. Note that as a result of this, the model diagnostic property contains lists instead of a scalars.
+
 0.5.2
 -----
 
--- a/ravenpy/config/commands.py
+++ b/ravenpy/config/commands.py
@@ -41,6 +41,20 @@ class LinearTransform(RavenCommand):
 
 
 @dataclass
+class EvaluationPeriod(RavenCommand):
+    """:EvaluationPeriod [period_name] [start yyyy-mm-dd] [end yyyy-mm-dd]"""
+
+    name: str
+    start: str
+    end: str
+
+    template = ":EvaluationPeriod {name} {start} {end}"
+
+    def to_rv(self):
+        return self.template.format(**asdict(self))
+
+
+@dataclass
 class SubBasinsCommand(RavenCommand):
     """SubBasins command (RVH)."""
 
--- a/ravenpy/config/commands.py
+++ b/ravenpy/config/commands.py
@@ -1,3 +1,4 @@
+import datetime as dt
 import itertools
 import re
 from abc import ABC, abstractmethod
@@ -45,8 +46,8 @@ class EvaluationPeriod(RavenCommand):
     """:EvaluationPeriod [period_name] [start yyyy-mm-dd] [end yyyy-mm-dd]"""
 
     name: str
-    start: str
-    end: str
+    start: dt.date
+    end: dt.date
 
     template = ":EvaluationPeriod {name} {start} {end}"
 
--- a/ravenpy/config/rvs.py
+++ b/ravenpy/config/rvs.py
@@ -814,6 +814,21 @@ class OST(RV):
     tmpl = """
     """
 
+    # Multiplier applied to metric before passing to minimization algorithm.
+    _evaluation_metrics_multiplier = dict(
+        NASH_SUTCLIFFE=-1,
+        LOG_NASH=-1,
+        RMSE=1,
+        PCT_BIAS=1,
+        ABSERR=1,
+        ABSMAX=1,
+        PDIFF=1,
+        TMVOL=1,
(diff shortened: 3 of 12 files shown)


huard:
@richardarsenault 

If we set EvaluationPeriod, the diagnostic file will include more lines, one for each period defined, plus the overall period. 

The Ostrich config related to this looks like:
```
BeginResponseVars
          #name   filename                              keyword         line    col     token
          NS      ./model/output/{run_name}-{run_index}_Diagnostics.csv;       OST_NULL        1       3       ','
        EndResponseVars

        BeginTiedRespVars
          NegNS 1 NS wsum -1.00
        EndTiedRespVars
```

If I understand correctly, you want to select the objective function to be used, and the period over which it is computed. With this PR, what will happen is that it will use by default `evaluation_metrics[0]`, and `evaluation_periods[0]`. Is this OK ?

Second issue is dealing with metrics that should not be multiplied by -1. Will that be an issue ? Suggestions?

richardarsenault:
@huard This looks good! I won't be able to actually test it before at least a few days, but looking at the code and tests it looks to be doing exactly what we need.

As for the minimization vs maximization: The issue is already present in the test since RMSE needs to be minimized directly while KGE needs to be maximized (so we take the negative value). I wonder if we should have some sort of flag? It might be cumbersome but then if people want to use custom metrics then they could simply add that flag and OSTRICH would know to multiply by -1 (or not). Not sure of the implementation difficulty though.

ravenpy/config/commands.py
```diff
@@ -40,6 +40,20 @@ def to_rv(self):
         return ""
 
 
+@dataclass
+class EvaluationPeriod(RavenCommand):
+    """:EvaluationPeriod [period_name] [start yyyy-mm-dd] [end yyyy-mm-dd]"""
+
+    name: str
```
cjauvin, line 47: Si tu utilises des `datetime.date` je crois que ça va faire ce que tu veux : accepter des arguments en string pour le constructeur (et les convertir), et convertir de nouveau en ISO au moment de formatter en string au render.

cjauvin:
```python
import datetime as dt

from pydantic.dataclasses import ValidationError, dataclass


@dataclass
class Bla:
    d: dt.date


b = Bla("2010-01-31")
print(b.d, type(b.d))

try:
    b2 = Bla("2010-01-32")
except ValidationError as e:
    print("got it even if it was a string!")

>>> 2010-01-31 <class 'datetime.date'>
got it even if it was a string!
```

ravenpy/config/rvs.py
```diff
@@ -386,6 +388,17 @@ def evaluation_metrics(self, values):
...
         self._evaluation_metrics = ms
 
+    @property
+    def evaluation_periods(self):
+        """:EvaluationPeriod option. Instantiate with list of EvaluationPeriod commands."""
+        return "\n".join([str(p) for p in self._evaluation_periods])
+
+    @evaluation_periods.setter
+    def evaluation_periods(self, values):
+        if not isinstance(values, (list, set, tuple)):
```
cjauvin, line 398: Je crois que j'utilise déjà `numpy.distutils.misc_util.is_sequence` dans ce même sous-module.

HISTORY.rst
```diff
@@ -2,6 +2,11 @@
 History
 =======
 
+0.6.0
+-----
+
+* Added support for EvaluationPeriod commands. Note that as a result of this, the model diagnostic property contains lists instead of a scalars.
```
cjauvin, line 8: ..the model's `diagnostics` property contains one list per key, instead of a single scalar.

cjauvin (approved):

ravenpy/config/commands.py
huard: Nice ! Done.

ravenpy/config/rvs.py
huard: Fait.

cjauvin:
Il faudrait fixer ce test je pense : `tests/test_blended.py::TestBLENDED_OST::test_simple`

huard:
blended semble avoir hardcodé `:EvaluationMetrics`, mais je pense qu'on l'ajoute ensuite RVI._post_tmpl

On se retrouve donc avec les metriques suivantes dans diagnostics: 
```
observed data series,filename,DIAG_NASH_SUTCLIFFE,DIAG_RMSE,DIAG_KLING_GUPTA,DIAG_NASH_SUTCLIFFE,DIAG_RMSE,
HYDROGRAPH_ALL,/home/david/src/raven-testdata/raven-gr4j-cemaneige/Salmon-River-Near-Prince-George_meteo_daily.nc,-1.47169,81.4569,-0.217328,-1.47169,81.4569
```
Bref, je pense qu'il faut enlever la section Output Options dans le template de blended.

huard:
@richardarsenault I added a `evaluation_metrics_multiplier` that is set to 1 or -1 depending on the case. Please check that my interpretation is correct.

richardarsenault:
@huard yes, interpretation is correct! Just PDIFF I am not sure exactly what it represents.

For PCT_BIAS, this is a special case because we need to target 0, so if we attempt to mimize it it could go to -100% bias and would be happy with that. We would need to take the absolute value and minimize that. Not sure how it's done internally...

huard:
PDIFF: peak difference. 

I don't think I see a way out for PCT_BIAS. I suggest we raise an Error if this is the metric used for Ostrich.

richardarsenault:
Question: Does OSTRICH already "know" that it has to bring PCT_BIAS to zero? like if that is coded already, and all we want it to display the correct value (min or max), then we can leave it at "1" and not take the negative, that would work perfectly.

huard:
No, my understanding is that Ostrich will try to minimize whatever metric we give it. It's our responsibility to pick a metric and a transformation. I don't see a transformation that would allow us to do this, and it seemed like an edge case that didn't warrant a huge effort.

PR #117, opened 2021-06-17
Support V2.1 of the Routing Product

cjauvin:
@huard @julemai @dustming This is the beginning of the effort to support V2.1 of the Routing Product. Please help me to validate everything, and come up with a variation on the `test_routing_lievre_tutorial` with which to test it.

### DIFF
--- a/tests/test_routing_lievre_tutorial.py
+++ b/tests/test_routing_lievre_tutorial.py
@@ -48,7 +48,7 @@ class TestRouting:
         # Model #
         #########
 
-        model = Raven("/tmp/ravenpy_debug/lievre/")
+        model = Raven()
 
         model.config.identifier = "raven-lievre-routing"
 
@@ -268,7 +268,7 @@ class TestRouting:
         # Model #
         #########
 
-        model = Raven("/tmp/ravenpy_debug/lievre_v21/")
+        model = Raven()
 
         model.config.identifier = "raven-lievre-routing-v21"
 
--- a/tests/test_routing_lievre_tutorial.py
+++ b/tests/test_routing_lievre_tutorial.py
@@ -223,9 +223,9 @@ class TestRouting:
         assert csv_lines[1].split(",")[:-1] == [
             "HYDROGRAPH_ALL",
             observation_data_nc_path.name,
-            "0.311049",
-            "-11.9514",
-            "0.471256",
+            "0.311049",  # NASH_SUTCLIFFE
+            "-11.9514",  # PCT_BIAS
+            "0.471256",  # KLING_GUPTA
         ]
 
         for d, q_sim in [
@@ -455,9 +455,9 @@ class TestRouting:
         assert csv_lines[1].split(",")[:-1] == [
             "HYDROGRAPH_ALL",
             observation_data_nc_path.name,
-            "0.253988",
-            "-17.0916",
-            "0.443201",
+            "0.253988",  # NASH_SUTCLIFFE
+            "-17.0916",  # PCT_BIAS
+            "0.443201",  # KLING_GUPTA
         ]
 
         for d, q_sim in [
--- a/tests/test_routing_lievre_tutorial.py
+++ b/tests/test_routing_lievre_tutorial.py
@@ -12,7 +12,7 @@ from ravenpy.extractors.routing_product import (
     RoutingProductGridWeightExtractor,
     RoutingProductShapefileExtractor,
 )
-from ravenpy.models import Raven
+from ravenpy.models import Raven, get_average_annual_runoff
 from ravenpy.utilities.testdata import get_local_testdata
 
 
@@ -138,7 +138,15 @@ class TestRouting:
         {channel_profiles}
         """
 
-        model.config.rvp.avg_annual_runoff = 594
+        total_area_in_m2 = sum(
+            [hru.area * 1000 * 1000 for hru in model.config.rvh.hrus]
+        )
+        model.config.rvp.avg_annual_runoff = get_average_annual_runoff(
+            observation_data_nc_path, total_area_in_m2, obs_var="Q"
+        )
+
+        assert model.config.rvp.avg_annual_runoff == pytest.approx(593.6191)
+
         model.config.rvp.soil_classes = [SoilClassesCommand.Record("AQUIFER")]
         model.config.rvp.soil_profiles = [
             SoilProfilesCommand.Record("Lake_Soil_Lake_HRU", ("AQUIFER",), (5,)),
@@ -229,7 +237,7 @@ class TestRouting:
         ]
 
         for d, q_sim in [
-            (0, 85.97869699520872),
+            (0, 85.92356056334998),
             (1000, 78.9516829670743),
             (2000, 70.29412760595676),
             (3000, 44.711237489482755),
@@ -369,7 +377,15 @@ class TestRouting:
         {channel_profiles}
         """
 
-        model.config.rvp.avg_annual_runoff = 594
+        total_area_in_m2 = sum(
+            [hru.area * 1000 * 1000 for hru in model.config.rvh.hrus]
+        )
+        model.config.rvp.avg_annual_runoff = get_average_annual_runoff(
+            observation_data_nc_path, total_area_in_m2, obs_var="Q"
+        )
+
+        assert model.config.rvp.avg_annual_runoff == pytest.approx(597.6287)
+
         model.config.rvp.soil_classes = [SoilClassesCommand.Record("AQUIFER")]
         model.config.rvp.soil_profiles = [
             SoilProfilesCommand.Record("LAKE", ("AQUIFER",), (5,)),
@@ -455,16 +471,16 @@ class TestRouting:
         assert csv_lines[1].split(",")[:-1] == [
             "HYDROGRAPH_ALL",
             observation_data_nc_path.name,
-            "0.253988",  # NASH_SUTCLIFFE
-            "-17.0916",  # PCT_BIAS
-            "0.443201",  # KLING_GUPTA
+            "0.253959",  # NASH_SUTCLIFFE
+            "-17.0904",  # PCT_BIAS
+            "0.443212",  # KLING_GUPTA
         ]
 
         for d, q_sim in [
-            (0, 85.401842),
-            (1000, 74.071142),
-            (2000, 62.679696),
-            (3000, 42.738307),
-            (4000, 128.689885),
+            (0, 85.92355875229545),
+            (1000, 74.05569855818379),
+            (2000, 62.675159400333115),
+            (3000, 42.73584909530037),
+            (4000, 128.70284018326998),
         ]:
             assert model.hydrograph.q_sim[d].item() == pytest.approx(q_sim)
--- a/tests/test_routing_lievre_tutorial.py
+++ b/tests/test_routing_lievre_tutorial.py
@@ -48,7 +48,7 @@ class TestRouting:
         # Model #
         #########
 
-        model = Raven()
+        model = Raven("/tmp/ravenpy_debug/lievre/")
 
         model.config.identifier = "raven-lievre-routing"
 
@@ -91,7 +91,9 @@ class TestRouting:
         #######
 
         rvh_extractor = RoutingProductShapefileExtractor(
-            routing_product_shp_path, hru_aspect_convention="ArcGIS"
+            routing_product_shp_path,
+            hru_aspect_convention="ArcGIS",
+            routing_product_version="1.0",
         )
         rvh_config = rvh_extractor.extract()
         channel_profiles = rvh_config.pop("channel_profiles")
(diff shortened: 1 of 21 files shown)


julemai:
This looks all very good to me. @dustming, could you just double-check the renaming of the variables.

dustming:
Just saw the message, got it.


Julie ***@***.***> 于2021年6月17日周四 下午9:55写道：

dustming:
The renaming of variables looks good to me too

cjauvin:
FYI, following a recent discussion with @julemai the next steps for this project will be:
1. Create a new RavenPy CLI script which will take two inputs: 
  (a) A V2.1 regional ("drainage") shapefile
  (b) The name of a gauge station
  and which will compute the subset of (a) that is "upstream" of (b).
2. Use (1) to generate a Lievre sample dataset which we can use to write a V2.1 variant of `test_lievre_tutorial`.

dustming:
> FYI, following a recent discussion with @julemai the next steps for this project will be:
> 
> 1. Create a new RavenPy CLI script which will take two inputs:
>    (a) A V2.1 regional ("drainage") shapefile
>    (b) The name of a gauge station
>    and which will compute the subset of (a) that is "upstream" of (b).
> 2. Use (1) to generate a Lievre sample dataset which we can use to write a V2.1 variant of `test_lievre_tutorial`.

Got it, Let me know if there is anything I can help with.  You may want to pay some attention to the "Use_region" attribute in the gauge shapefile.  When it is 1, we need to get the subset of subbasins for that gauge from the drainage region product; when it is zero, we can get the subset of the subbasins for that gauge from sub drainage region product.

julemai:
Hi @dustming!

I think it would be very helpful if you could give example filenames for the three shape files you talk about, i.e.

- gauge shapefile
- drainage region product
- sub drainage region product

Thanks!

dustming:
> Hi @dustming!
> 
> I think it would be very helpful if you could give example filenames for the three shape files you talk about, i.e.
> 
> * gauge shapefile
> * drainage region product
> * sub drainage region product
> 
> Thanks!

gauge shapefile:  obs_gauges_v2-1.shp
drainage region product: for example shapefiles in drainage_region_0004_v2-1
sub drainage region product: for example shapefiles in drainage_region_0004_00015_v2-1

We do not need to consider this if you use the product in drainage_region_0_v2-1.

julemai:
Thanks Ming! We are planning to use only the "regional" products (I guess that is what you call "drainage region product" above). So, I assume we won't need to look anything up in the gauge shapefile, right?

dustming:
Oh, I see, yes, you do not need to look into gauge shapefile in this case.

cjauvin:
@julemai @huard @dustming The PR is now ready to be reviewed. I have created a new script  we talked with Julie (included in this PR), used it on `drainage_region_0003_v2-1/finalcat_info_v2-1.shp` with gauge `02LE024` to create `lievre_subbasins_v21.shp`. I have then used Ming's script to generate `lievre_hrus_v21.shp`, which is avaiable at https://github.com/Ouranosinc/raven-testdata/blob/master/raven-routing-sample/lievre_hrus_v21.zip.

I have created a new test `test_lievre_tutorial_v21` that makes use of it. Please validate everything I'm doing in this new test (the gauge manipulations, the updates to the numerical testing stuff, in relation to the corresponding V1 test, etc). Thanks for your collaboration.

tests/test_routing_lievre_tutorial.py
```diff
@@ -234,3 +236,235 @@ def test_lievre_tutorial(self):
...
+        :PrecipIceptFract      PRECIP_ICEPT_NONE             # Estimation of the precipitation interception fraction. In this routing model, stream input(s) are "pretending" to be precipitation going into Raven, thus using DEFAULT PRECIP_ICEPT_NONE to indicate no interception processes are adopted.
+        :PotentialMeltMethod   POTMELT_NONE                  # Estimation of the potential snow melt. In this routing model, snow melt processes are not relevant, thus using DEFAULT POTMELT_NONE method.
+        :SoilModel             SOIL_ONE_LAYER                # In this routing model, use DEFAULT SOIL_ONE_LAYER to define single soil layer structure.
+
+        :HydrologicProcesses
+          :Precipitation     PRECIP_RAVEN             ATMOS_PRECIP     PONDED_WATER          # Moves stream input(s) from ATMOS_PRECIP to PONDED_WATER storage (waiting for runoff). Use DEFAULT PRECIP_RAVEN method.
+          :Flush             RAVEN_DEFAULT            PONDED_WATER     SURFACE_WATER         # Moves water from PONDED_WATER to SURFACE_WATER (routed to outlet). Use DEFAULT RAVEN_DEFAULT method.
+        :EndHydrologicProcesses
+        """
```
huard, line 290: Could this be replaced with the Routing emulator ?

huard (approved):

tests/test_routing_lievre_tutorial.py
cjauvin: I'm not sure what you mean as I have removed the routing emu a while ago.. it wasn't really an emulator per se no?

huard: I hadn't realized this. You're right that it's a bit different to other emulators, but I thought it was nice to have it around.

cjauvin: I did not understood in what sense it was actually an emulator.. it didn't have any particular parameter, etc. If you want it back I'm happy to discuss it though.

huard: I don't have strong feelings about it. It may have made the test code here a bit simpler, but I think it may be confusing to users if there's an emulator that is substantially different from others. Let's leave it at that for now.

tests/test_routing_lievre_tutorial.py
```diff
@@ -234,3 +236,235 @@ def test_lievre_tutorial(self):
...
+        {vegetation_classes}
+
+        {land_use_classes}
+
+        {avg_annual_runoff}
+
+        {channel_profiles}
+        """
+
+        model.config.rvp.avg_annual_runoff = 594
```
julemai, line 372: I am happy with this being hard-coded here. But we had a function to derive those values, right?

tests/test_routing_lievre_tutorial.py
```diff
@@ -234,3 +236,235 @@ def test_lievre_tutorial(self):
...
+        assert len(model.hydrograph.time) == (end - start).days + 1
+
+        assert model.hydrograph.basin_name.item() == gauged_sb.name
+
+        csv_lines = model.outputs["diagnostics"].read_text().split("\n")
+
+        assert csv_lines[1].split(",")[:-1] == [
+            "HYDROGRAPH_ALL",
+            observation_data_nc_path.name,
+            "0.253988",
```
julemai, line 458: Shall we add comments here that these are:
```
"NASH_SUTCLIFFE",
"PCT_BIAS",
"KLING_GUPTA"
```

tests/test_routing_lievre_tutorial.py
```diff
@@ -234,3 +236,235 @@ def test_lievre_tutorial(self):
...
+            "0.443201",
+        ]
+
+        for d, q_sim in [
+            (0, 85.401842),
+            (1000, 74.071142),
+            (2000, 62.679696),
+            (3000, 42.738307),
+            (4000, 128.689885),
+        ]:
```
julemai, line 469: I am very happy with how similar those values are compared to the v1.

julemai:
Hi @cjauvin! 

I went through the changed files and the setup created for v2.1. I made some comments. Nothing serious. Here are some general comments:
- Would it be possible to have a header in the five RV* files that states that this is a Raven setup file, maybe version, maybe date created, maybe stating who initially provided these setups, etc...?
- In the currently produced RVH file there is one line in the `SubBasins` block that is:
```
3077541    sub_3077541 -1         chn_3077541 2.99255    1          02LE024 
```
but should be 
```
3077541    sub_3077541 -1         chn_3077541 2.99255    1          # 02LE024 
```
- I am wondering if we should also provide a Ostrich setup for this such that one could calibrate the routing model?!

All the rest looks great to me. Good job!!

julemai:
One other application that comes to mind... Not sure if this is the right place to say...

The user could want to include only lakes above a certain size (let's say 1 km2) in his/her model. So, even if the routing product says that a basin contains a lake we would ignore this and handle the sub basin as a land-only. It would be an optional argument. The default would be to include lakes with a lake area above 0 km2 (all lakes).

cjauvin:
> I am wondering if we should also provide a Ostrich setup for this such that one could calibrate the routing model?!

Up to you and @huard to provide an answer (and additional code if the answer is yes!) for that.

tests/test_routing_lievre_tutorial.py
cjauvin: done

julemai:
Re Ostrich setup: The only two parameters to calibrate are the following two in the RVH file:

```
:SBGroupPropertyMultiplier Land  MANNINGS_N            1.0
:SBGroupPropertyMultiplier Lakes RESERVOIR_CREST_WIDTH 1.0
```
as
```
:SBGroupPropertyMultiplier Land  MANNINGS_N            par_x01
:SBGroupPropertyMultiplier Lakes RESERVOIR_CREST_WIDTH par_x02
```

with ranges being `par_x01` in [0.2,5.0] and `par_x02` in [0.1,2.0].

dustming:
> One other application that comes to mind... Not sure if this is the right place to say...
> 
> The user could want to include only lakes above a certain size (let's say 1 km2) in his/her model. So, even if the routing product says that a basin contains a lake we would ignore this and handle the sub basin as a land-only. It would be an optional argument. The default would be to include lakes with a lake area above 0 km2 (all lakes).

Here is a potential problem to remove lakes in this step.  
    a) For all lake subbasins, the river length is zero in the HRU polygon and the finalcat_info_v2-1.shp polygon. 
    b) If we may lose some routing information by saying Lake subbasin does not have a lake.  
Here is a potential workflow:
    a) Use the BasinMaker post-processing function to remove lakes first. And then, we will obtain a simplified routing network.
    b) use the script to generate HRU polygon from the simplified routing network. 
    c) and then generate raven input files.

tests/test_routing_lievre_tutorial.py
cjauvin: We do (`get_average_annual_runoff`) and I updated the two Lievre tests so that they now make use of it.

cjauvin:
> In the currently produced RVH file there is one line in the SubBasins block that is:

@julemai fixed

tests/test_routing_lievre_tutorial.py
julemai: Nice!!

julemai:
> > One other application that comes to mind... Not sure if this is the right place to say...
> > The user could want to include only lakes above a certain size (let's say 1 km2) in his/her model. So, even if the routing product says that a basin contains a lake we would ignore this and handle the sub basin as a land-only. It would be an optional argument. The default would be to include lakes with a lake area above 0 km2 (all lakes).
> 
> Here is a potential problem to remove lakes in this step.
> a) For all lake subbasins, the river length is zero in the HRU polygon and the finalcat_info_v2-1.shp polygon.
> b) We may lose some routing information by saying Lake subbasin does not have a lake.
> Here is a potential workflow:
> a) Use the BasinMaker post-processing function to remove lakes first. And then, we will obtain a simplified routing network.
> b) use the script to generate HRU polygon from the simplified routing network.
> c) and then generate raven input files.

Thanks @dustming for your reply. Unfortunately, making use of the BasinMaker is not really an option since we would need to use the "post-processing" option on the fly when the user specifies the lake area threshold. We are not currently planning to include the BasinMaker- especially not in the full installation since it is too complex to get it supported and installed on all platforms etc.

I will follow up with you and Bryan/James on Monday re this. In the "worst" case we just won't support the lake threshold. It was just an idea that came to my mind.

dustming:
No problem, just want to point out the potential problem to simply
remove lakes in Lakes.rvh file.

Cheers
Ming

Julie ***@***.***> 于2021年7月10日周六 下午7:47写道：

julemai:
Alright. Coming back to my idea of allowing users to specify a threshold of lakes to be included: As @dustming mentioned there is an issue that the river length is missing for basins that contain lakes. I was initially thinking that this would be available in the routing product but it apparently is not. So, let's forget that I mentioned this feature (for now). It might be included in a later version of the routing product but currently it isn't and it would be too much work to include.

@cjauvin, meaning that including the header in the created files is the last thing to address. :)

cjauvin:
@julemai All right! Here's the requested common RV template header mechanism: https://github.com/CSHS-CWRA/RavenPy/pull/117/commits/9da8faa60b4e9596195954dbb65697b5e4501ae3

I took the occasion to refactor some related internal stuff (mainly in `models.base` and `models.config.rvs`), to make it more coherent.

Last question before we can merge: should we add @dustming's HRU generation script to RavenPy, as @huard suggested?

cjauvin:
@julemai This is the way to add a `description` to the header: https://github.com/CSHS-CWRA/RavenPy/blob/f67462e8df95a87cc9d6894f319959cfe336e2ab/tests/test_routing_lievre_tutorial.py#L280-L283

julemai:
@cjauvin, excellent regarding the header. 

Re your question with HRU generation. I talked with @richardarsenault the other day and he said there is already logic of how to derive HRUs in PAVICS. I think it is better to use that logic to define HRUs for the land-part of each subbasin through that existing code. But it would definitely be a new PR because the HRU definition can be applied to ANY model not only the distributed versions (a.k.a the setup where we use the routing product).

julemai (approved):

cjauvin:
Thanks a lot to everyone, I will then merge these changes soon.

huard:
@richardarsenault Could you elaborate? I'm not sure I follow.

richardarsenault:
The idea we had discussed a while back during our first forays into distributed modelling was to use the HydroBasins level 12 subbasins that could be considered as HRUs. We already have the logic to find these HRUs within a catchment (as done in the catchment delineation toolbox) and then we can extract the land cover/etc. for each of these HRUs. All that is missing, I think, is to tie these things together.

That is my understanding, at least. Maybe @Zeitsperre can chime in for the HRU identification?

huard:
Ok, got it. We'll need to open a ticket to implement the full workflow logic though.

Zeitsperre:
You're right, @richardarsenault, that's more or less the approach we took for the HydroBASINS data and the same workflow needs to be chained together for the Routing Product. Finding upstream and related HRUs using Pandas-logic is the final piece of the puzzle, I think.

PR #118, opened 2021-07-08
Add point intersection calls

Zeitsperre:


Zeitsperre:
@huard I'll need to make a few adjustments in the RavenWPS code, but everything is working as expected. Nicely done!

huard (approved):
Excellent !

huard:
@Zeitsperre Any idea why the conda build has errors ?

Zeitsperre:
Not yet, but I'm gearing up to look into that this morning, though it's a nice surprise to see that the PyPI-only builds are working as expected. Will post updates on slack.

PR #119, opened 2021-07-14
Owslib compatibility fixes

Zeitsperre:
This is a transitional set of fixes for backwards compatibility. Be aware that tests using `owslib<=0.24.1` and a newer version of GeoServer will fail.

### DIFF
--- a/tests/test_geoserver.py
+++ b/tests/test_geoserver.py
@@ -7,6 +7,8 @@ from ravenpy.utilities.testdata import get_local_testdata
 
 pytestmark = pytest.mark.online
 
+# FIXME: Remove XFAIL marks once OWSLib > 0.24.1 is released.
+
 
 class TestHydroBASINS:
     geoserver = pytest.importorskip("ravenpy.utilities.geoserver")
@@ -25,6 +27,7 @@ class TestHydroBASINS:
         dom = self.geoserver.select_hybas_domain(bbox)
         assert dom == "ar"
 
+    @pytest.mark.xfail(reason="OWSLib>0.24.1 is needed.")
     def test_get_hydrobasins_location_wfs(self, tmp_path):
         lake_winnipeg = (
             -98.03575958286369,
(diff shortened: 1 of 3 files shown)


tests/test_geoserver.py
```diff
@@ -25,6 +27,7 @@ def test_select_hybas_ar_domain(self):
         dom = self.geoserver.select_hybas_domain(bbox)
         assert dom == "ar"
 
+    @pytest.mark.xfail(reason="OWSLib>0.24.1 is needed.")
```
huard, line 30: Not sure I understand. I think we probably want those tests to xfail if owslib < 0.24.1. As it is, it will always fail. 
e.g. pytest.importorskip with minversion. 
https://docs.pytest.org/en/6.2.x/reference.html#pytest-importorskip

Zeitsperre: I can set it to firmly need the newest OWSLib, but that would break the current version, no? The tests are now written to support the old (≤0.24.1) and new (>0.24.1) OWSLib, but will absolutely fail if we update the GeoServer without updating OWSLib (to >0.24.1). We can't pin a `git+https`-installed package in conda-forge (nor do we want to).

The reason why I took this approach is that I can't find the release schedule for OWSLib, and I'd rather have something that will remain functional at least until the next release.

As it stands, the only XFAIL-ing test is the direct call to `_get_location_wfs` using `point` and `owslib<=0.24.1`. If this doesn't make sense as an approach, let me know.

huard: I agree with the intent, but but where does the test check the version of owslib ? That's the part I'm missing.

Zeitsperre: The "check" is in the code of the library itself, since any workaround to support the older OWSLib version would be needed at that level of proximity. I think the consideration I took here are:
1. The current test ensemble should not be broken with older/newer owslib
2. Users should not need to modify their scripts/calls
3. RavenWPS function should use the newer API (reading json-encoded data, not bytestrings)
4. migrating to the newer OWSlib should be as painless as possible (the only file that needs to be modified is `geoserver.py`).

I think I understand where you're coming from. These tests aren't written to fail with older/newer owslib, but only IFF the older owslib is used with the newer geoserver. If it makes sense, I'll update the code of the library to handle just this particular case.

huard: Ok, I missed this nuance in your PR. If the tests are still exercised for other combinations, it's good on my end.

huard (approved):

PR #121, opened 2021-07-21
Remove clisops

Zeitsperre:
We use CLISOPS for precisely one subsetting operation that can be replicated using xarray and rioxarray. CLISOPS is quite heavy on dependencies and has a few strange behaviours, like very aggressive logging (see: https://github.com/roocs/clisops/issues/144), that we don't need. Removing it means removing many libraries we don't already use.

### DIFF
--- a/.cruft.json
+++ b/.cruft.json
@@ -10,7 +10,7 @@
             "project_slug": "ravenpy",
             "project_short_description": "A Python wrapper to setup and run the hydrologic modelling framework Raven.",
             "pypi_username": "CSHS-CWRA",
-            "version": "0.6.0",
+            "version": "0.6.1",
             "use_pytest": "y",
             "use_pypi_deployment_with_travis": "y",
             "add_pyup_badge": "y",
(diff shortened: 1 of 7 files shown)


cjauvin (approved):

.cruft.json
```diff
@@ -10,7 +10,7 @@
             "project_slug": "ravenpy",
             "project_short_description": "A Python wrapper to setup and run the hydrologic modelling framework Raven.",
             "pypi_username": "CSHS-CWRA",
-            "version": "0.6.0",
+            "version": "0.6.1",
```
cjauvin, line 13: with all the changes that have been merged recently, I think that bumping the minor would probably make more sense though

cjauvin (approved):

.cruft.json
Zeitsperre: Bumping to 0.6.1 for debugging purposes. There are a handful of changes needed before we release a next minor version. Right now, the master branch relies on some git@master versions, and I'd like to sanitize that before we release a new version. This requires a game plan, so let's discuss over slack.

Issue #125, opened 2021-07-22
Moving logic from `ravenpy.utilities` to other packages (as much as possible)


cjauvin:
In order to remove heavy dependencies and further reduce the conceptual surface covered by RavenPy (its core mission, after all, is to be essentially a RV config generator, along with a Raven process launcher and manager), we (@cjauvin and @Zeitsperre) have discussed today the possibility of moving logic away from `ravenpy.utilities` as much as possible: some components (like `geoserver`) should maybe go back to RavenWPS, and some others into a new separate package, with a name like ~RavenGIS (from which RavenWPS would depend, but not RavenPy).

Zeitsperre:
Moving GIS functions out of there would not only help set the scope of RavenPy, but would also make maintenance of the GIS dependencies much simpler to manage. It would also allow us to remove the GIS warnings from many modules. It wouldn't require much effort either (`cookiecutter` would do 90% of the work). Seems like a slam dunk.

PR #127, opened 2021-08-05
RavenGIS

Zeitsperre:
## Overview

This PR fixes [issue id]

Changes:

* Refactoring of the GIS utilities found in RavenPy into its own project

## Related Issue / Discussion

https://github.com/CSHS-CWRA/RavenPy/issues/125

## Additional Information

Links to other issues or sources.


ravenpy/extractors/routing_product.py
```diff
@@ -359,35 +360,69 @@ def __init__(
         self._input_is_netcdf = True
 
         input_file_path = Path(input_file_path)
-        if input_file_path.suffix == ".nc":
+        if {input_file_path.suffix.lower()}.issubset({".nc", ".nc4"}):
```
cjauvin, line 363: You can use `in`: 

```python
if input_file_path.suffix.lower() in {".nc", ".nc4"}:
```

Zeitsperre:
Present state of affairs: Tests are passing, but the speed of tests `test_generate_grid_weights_with_shp_input` and `test_generate_grid_weights_with_weight_rescaling` is really slow. I imagine there are some optimizations that need to be worked out. Besides those, speed is comparable to GDAL implementations. Will begin cleaning up the PR here now.

Zeitsperre:
There's a very real possibility that the code will run faster without GDAL. I have some optimizing left to do, but I think the slowdowns in the GridWeight file formatter can be fixed with a cleaner approach. The original version of the code is triggered automatically if `geopandas` is installed (which depends on GDAL).

Zeitsperre:
I can't explain why the PyPI-noGDAL-based install is incredibly slow on some tests. The conda-forge build does not have GDAL and outperforms PyPI-GDAL builds by minutes. In any case, I need to move on for now,

Issue #132, opened 2021-08-30
Adapt to new SDBA object methods (xclim v0.29.0)


Zeitsperre:
* RavenPy version: master (v0.14b0)
* Python version: 3.9
* Operating System: Linux

The newest Xclim (v0.29.0) changed the way that SDBA objects are used. The API needs adjustment here.

PR #133, opened 2021-08-30
Temporarily pin xclim below 0.29.0

Zeitsperre:
Opening this PR to give us some wiggle room in case #132 requires a fair amount of effort.


huard (approved):
In any case, I think it's good that we do not require RavenPy users to update so soon to the new xclim.

PR #136, opened 2021-09-07
Fix test with xclim 0.29

huard:


tests/test_bias_correction.py
```diff
@@ -1,4 +1,5 @@
 import xarray as xr
+import xclim.sdba
 import xclim.sdba as sdba
```
Zeitsperre, line 3: ?

Zeitsperre (approved):

tests/test_bias_correction.py
huard: TBH, I don't even know why this test is in RavenPy... it's essentially testing xclim, so apart from giving some pointers on how to use sdba, this is not very useful. I suspect this is a building block for future stuff, but let's not spend too much time on this for now.

huard:
I'll wait before merging and make sure Raven is working as expected.

tests/test_bias_correction.py
Zeitsperre: I imagine we could think of this test as a sanity check for the xclim API in Ravenpy. We do  similar tests in xclim for CLISOPS to ensure that the core functions we rely on are still working as expected.

PR #140, opened 2021-10-13
Update with coveralls via cookiecutter

Zeitsperre:


.coveralls.yml
```diff
@@ -0,0 +1,2 @@
+service_name: github
+repo_token: YOURTOKENHERE
```
huard, line 2: I guess we need to fill this, no ?

huard (approved):

.coveralls.yml
Zeitsperre: Thankfully, no. If we want to enable coveralls, we can leverage GitHub Actions directly. The case where we would need this is if we wanted to report to Coveralls manually.

Zeitsperre:
With all the tox-related code implemented already in here, I think the second step for this PR is to add the GitHub Action for reporting coverage. I can get to this next week. For an example of what that looks like, see: https://github.com/Ouranosinc/xclim/blob/d5e7eb984af99c0f6696e9678c18085fa3b551db/.github/workflows/main.yml#L64 (line 64 to end of script).

cjauvin (approved):

Zeitsperre:
Found the source of the failures(https://github.com/Toblerity/Shapely/releases/tag/1.8.0):
> The deprecated to_wkb and to_wkt methods of BaseGeometry have been removed.

Pinning shapely at 1.7.1 for now. Will see if that works.

PR #147, opened 2021-12-20
Fix assimilation

richardarsenault:
Data assimilation had a small error which made it unsusable. 

The problem was an index on the time variables which was constant instead of advancing in time. 

This is now fixed, as is the test file.

I would want to merge this so I can work on the notebook in RavenWPS.


huard (approved):

Zeitsperre:
There's an issue with the VM environment on GitHub. The fix is currently in #146, so I'll merge that first.

PR #149, opened 2021-12-20
Open pinned gis libraries and update calls to remove shapely `to_wkt()` calls

Zeitsperre:


cjauvin (approved):

environment.yml
```diff
@@ -25,14 +25,14 @@ dependencies:
...
-  - rasterio <=1.2.6
+  - pyproj >=3.0
+  - rasterio
   - requests
   - rioxarray
   - scikit-learn ==0.24.2
   - scipy
   - setuptools >50.0
-  - shapely <=1.7.1
+  - shapely
```
tlvu, line 35: Oh @Zeitsperre , you'll release another RavenPy to conda? !

Zeitsperre: That;'s the idea. I wanted to ensure that the changes I've already pushed work currently. I can tag and publish today if everything is working.

PR #150, opened 2021-12-21
Add functionalities in Data Assimilation utils and simplify tests

richardarsenault:
The Data assimilation utilities were relatively simple and required users to setup long chains of processes to perform. The test case was particularly long, and had lots of direct modifications of the Raven model instance, in a complex loop.

I simplified this by adding functions in the data_assimilation utils that does this work for the user. The test-case has been simplified as well to make use of these tools. Now users can call the DA tools easily through Notebooks and ravenpy utils directly.

Next step will be to add a notebook to display the usage, which will come soon in the Raven-WPS repo.

### DIFF
--- a/ravenpy/utilities/data_assimilation.py
+++ b/ravenpy/utilities/data_assimilation.py
@@ -243,11 +243,7 @@ def perturb_full_series(model, std, start_date, end_date, dists, n_members=25):
                 member=n_members,
             )
 
-            # Save flow for later
-            if key == "water_volume_transport_in_river_channel":
-                q_obs = da
-
-    return perturbed, q_obs
+    return perturbed
 
 
 def assimilation_initialization(
--- a/ravenpy/utilities/data_assimilation.py
+++ b/ravenpy/utilities/data_assimilation.py
@@ -198,20 +198,19 @@ def update_state(x, qobs_pert, qobs_error, qsim):
 
 def perturb_full_series(model, std, start_date, end_date, dists, n_members=25):
     """
-    This function will take a timeseries as input and perturb each variable
-    according to the desired uncertainty/confidence level and distribution of
-    each variable. Returns an Xarray Dataset object with n_members x time values
-    for each variable.
+    Create an ensemble of randomly perturbed input timeseries for the given model.
+    Each input variable is modified by 'n_members' time series of random values
+    drawn from a given distribution and parameters.
 
     Parameters
     ----------
-    model : Raven model class object
+    model : ravenpy.Raven instance
         The model that will be used to perform the simulations and assimilation
-    std : tuple
-        Pairs of variable: standard deviation of the uncertainty.
-    start_date : datetime object
+    std : dict
+          Standard deviation of the perturbation noise for each input variable, keyed by variable standard name.
+    start_date : datetime
         Start date of the assimilation run.
-    end_date : datetime object
+    end_date : datetime
         End date of the assimilation run.
     dists : tuple
         Pairs of variable:distribution to identify the type of distribution each variable follows.
@@ -220,8 +219,8 @@ def perturb_full_series(model, std, start_date, end_date, dists, n_members=25):
 
     Returns
     -------
-    perturbed: path object (string)
-        Path to the perturbed hydrometeorological timeseries for the entire period.
+    perturbed: xarray.Dataset
+        Dataset containing all perturbed hydrometeorological timeseries for the entire period.
     """
 
     # Use the same random seed for both tasmin and tasmax
@@ -248,12 +247,7 @@ def perturb_full_series(model, std, start_date, end_date, dists, n_members=25):
             if key == "water_volume_transport_in_river_channel":
                 q_obs = da
 
-    # Create netcdf for the model.
-    p_fn = model.workdir / "perturbed_forcing.nc"
-    perturbed = xr.Dataset(perturbed)
-    perturbed.to_netcdf(p_fn, mode="w")
-
-    return p_fn, q_obs
+    return perturbed, q_obs
 
 
 def assimilation_initialization(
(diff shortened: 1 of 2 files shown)


ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
 
     return xa
+
+
+def perturb_full_series(model, std, start_date, end_date, dists, n_members=25):
+    """
+    This function will take a timeseries as input and perturb each variable
+    according to the desired uncertainty/confidence level and distribution of
+    each variable. Returns an Xarray Dataset object with n_members x time values
+    for each variable.
```
huard, line 204:
```suggestion
    """Create an ensemble of randomly perturbed input timeseries for the given model.
    
    Each input variable is modified by `n_members` time series of random values drawn from a 
    given distribution and  parameters. 
```

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+
+    Parameters
+    ----------
+    model : Raven model class object
+        The model that will be used to perform the simulations and assimilation
+    std : tuple
+        Pairs of variable: standard deviation of the uncertainty.
+    start_date : datetime object
+        Start date of the assimilation run.
+    end_date : datetime object
```
huard, line 214:
```suggestion
    end_date : datetime
```

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    according to the desired uncertainty/confidence level and distribution of
+    each variable. Returns an Xarray Dataset object with n_members x time values
+    for each variable.
+
+    Parameters
+    ----------
+    model : Raven model class object
+        The model that will be used to perform the simulations and assimilation
+    std : tuple
+        Pairs of variable: standard deviation of the uncertainty.
```
huard, line 211: How does this work if dist is not norm ? There is always just one parameter ?
Maybe a dict would be a better type here than a tuple.

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    start_date : datetime object
+        Start date of the assimilation run.
+    end_date : datetime object
+        End date of the assimilation run.
+    dists : tuple
+        Pairs of variable:distribution to identify the type of distribution each variable follows.
```
huard, line 217: Same here, I suggest to use a dict.

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    n_members : integer
+        Number of members to use in the Ensemble Kalman Filter.
+
+    Returns
+    -------
+    perturbed: path object (string)
+        Path to the perturbed hydrometeorological timeseries for the entire period.
```
huard, line 224: I would suggest returning the dataset, instead of writing to disk. I suggest the writing to disk should be done one level up, e.g. by the assimilation workflow.

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+                da,
+                dists.get(key, "norm"),
+                std=s,
+                seed=seed.get(key, None),
+                member=n_members,
+            )
+
+            # Save flow for later
+            if key == "water_volume_transport_in_river_channel":
+                q_obs = da
```
huard, line 249: This is a bit weird.

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    end_date,
+    area,
+    elevation,
+    latitude,
+    longitude,
+    params,
+    assim_var,
+    n_members=25,
+):
+
```
huard, line 272: Add docstring

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    basin_states,
+    perturbed_dataset,
+    q_obs,
+    assim_var,
+    start_date,
+    end_date,
+    n_members=25,
+    assim_step_days=7,
+):
+
```
huard, line 329: docstring

ravenpy/utilities/data_assimilation.py
```diff
@@ -191,3 +194,173 @@ def update_state(x, qobs_pert, qobs_error, qsim):
...
+    # ==== Assimilation ====
+    q_assim = []
+    sd = start_date
+
+    # Make list of days on which we will assimilate
+    number_days = end_date - start_date
+    number_days = number_days.days
+    assim_dates = math.floor(number_days / assim_step_days) * [assim_step_days]
+
+    for i, ndays in enumerate(assim_dates):
```
huard, line 339: @richardarsenault You're writing nice python code !

ravenpy/utilities/data_assimilation.py
richardarsenault: it seems it is already a dictionnary, I just mixed up the terms because I didn't define it using "dict(...)" but it seems I am still learning :)

ravenpy/utilities/data_assimilation.py
richardarsenault: Thanks, but most of it I copied from the test_data_assimilation function that YOU wrote so.... yeah... you write nice python code! :)

ravenpy/utilities/data_assimilation.py
richardarsenault: I agree, and I am not too sure how to handle this. At one point we "force" users to use this key to make sure there is Qobs in the netcdf. But I see here that the variable name we are using is "qobs" in our netcdf file, so I am guessing this is the long_name? But I can't seem to access the array by long_name, just the variable name. So I am a bit stumped on how to make this better. Ideas?

richardarsenault:
@huard Thanks for the review. I corrected pretty much everything except the q_obs key issue, if you have any ideas that'd be great!

ravenpy/utilities/data_assimilation.py
huard: No, what I'm confused about is that we return the perturbed dataset, including flow, + the original flow. I think the code would be cleaner if we just returned perturbed.

richardarsenault:
OK I've removed the qobs part and now extract it independently from the parent function. Users don't actually need to have it unless they want to compute skill scores or plot results, so I agree it needed to move out of the perturbation function.

PR #153, opened 2022-01-04
Fixes for notebooks

cjauvin:
These are changes for things I am discovering in the context of working with RavenPy in a shared notebook environment (not PAVICS, to be clear) with Aida (@Idajad):

* If your input files are in the notebook environment, the symlink to the NC input was not working
* I also think it's better to make the Raven warnings more visible to the user

### DIFF
--- a/ravenpy/models/base.py
+++ b/ravenpy/models/base.py
@@ -21,6 +21,7 @@ from collections import OrderedDict
 from dataclasses import astuple, fields, is_dataclass, replace
 from pathlib import Path
 from typing import Any, Dict, List, Union, cast
+from warnings import warn
 
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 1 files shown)


ravenpy/models/base.py
```diff
@@ -267,7 +277,7 @@ def setup_model_run(self, ts):
         # Create symbolic link to input files
         for fn in ts:
             if not (self.model_path / Path(fn).name).exists():
-                os.symlink(str(fn), str(self.model_path / Path(fn).name))
+                os.symlink(os.path.realpath(fn), str(self.model_path / Path(fn).name))
```
Zeitsperre, line 280: Did you want this to be a strict check for a real path? ie: https://docs.python.org/3/library/os.path.html#os.path.realpath

Zeitsperre (approved):

ravenpy/models/base.py
cjauvin: Not sure I understand your question. The problem was that the "source" part of the symlink was not good (in the case that the file is coming from the notebook env) so I replaced it (the source) with the full absolute path. The "destination" stays the same.

Zeitsperre: Ah, I understand now. I also had no idea that `realpath` expanded symlinks. Very useful. (https://stackoverflow.com/a/40311142/7322852)

huard (approved):
Joli

PR #174, opened 2022-04-08
Aida famine - do not merge

huard:


tests/test_commands.py
```diff
@@ -162,7 +165,7 @@ def test_tied_params():
         """
         BeginTiedParams
           ex_1 1 X1 linear 0.2 0 free
-          ex_2 2 X1 X2 linear 0 2 3 10 free
+          ex_2 2 X1 X2 linear 0 3 2 10 free
```
cjauvin, line 168: @huard Ce fix est nécessaire pour faire passer le test mais je trouve l'inversion des coefficients 2 et 3 suspecte (je suis pas certain de bien interpréter la syntaxe par contre).. peux-tu valider stp?

tests/test-Mohammad.py
```diff
@@ -0,0 +1,66 @@
...
+from warnings import warn
+
+import numpy as np
+import xarray as xr
+from numpy.distutils.misc_util import is_sequence
+
+from ravenpy.extractors.routing_product import RoutingProductShapefileExtractor
+from ravenpy.models import Raven
+
+model = Raven(identifier="famine", workdir="./test_rv_Mohammad")
```
cjauvin, line 66: @huard @Bijan55699 Not sure why this is there but I will remove it.

ravenpy/extractors/routing_product.py
```diff
@@ -116,19 +118,21 @@ def extract(self, model=None):
                 reservoir_cmds.append(self._extract_reservoir(row))
                 is_lake = True
             elif row[lake_field] > 0:
-                continue
+                pass  # not sure about this condition!
```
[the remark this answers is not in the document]
cjauvin: finalement je le remets comme avant, case closed..

tests/test_commands.py
huard: Non, c'est beau. La version initiale du code avait les coefficients à l'envers. J'ai corrigé dans le code, mais pas mis à jour le test.

cjauvin:
Also I see we have some tests that failed because the ouranos network was down I think, this is not good.

cjauvin:
> @Zeitsperre If you have time could you please take a look at the numpy error with the conda setup. I tried to reproduce it locally by upgrading to numpy 1.23, to no avail. Not sure what goes on there.

@Zeitsperre I see the commit you last pushed is green, did you fix this particular issue? I cannot see what you did specifically.

Zeitsperre:
There are two issues, sadly, but the fix is in this commit: https://github.com/CSHS-CWRA/RavenPy/commit/70517b995eec87e38fc28f03392335e16a435701

NumPy and Python-NetCDF4 had recent releases. These releases have issues:

1. NumPy v1.23 has changed some logic handling and this causes two test failures:
```
FAILED tests/test_cli.py::TestAggregateForcingsToHRUs::test_aggregate_forcings_to_hrus
FAILED tests/test_cli.py::TestAggregateForcingsToHRUs::test_aggregate_forcings_to_hrus_with_nodata
```
In order to address this, NumPy is pinned at v1.22 for the moment. The second issue is weirder:

2. Python-NetCDF4 v1.6.0 behaves differently depending on whether it's been installed via pip or Anaconda. The Anaconda binaries seem to be perfectly fine, but with the pip library we get errors like this: https://github.com/CSHS-CWRA/RavenPy/runs/7617722181?check_suite_focus=true#step:6:846

The good news is that the NumPy issue can likely be fixed pretty easily by updating the internals of the `test_aggregate_forcings_to_hrus*` tests. Not sure how to proceed with Python-NetCDF4, but here's a similar issue on their tracker: https://github.com/Unidata/netcdf4-python/issues/1179

cjauvin:
> The good news is that the NumPy issue can likely be fixed pretty easily by updating the internals 

That was my original intention, but the problem I have is that I could not reproduce the issue locally, using Numpy 1.23.. I really don't understand why!

Zeitsperre:
@cjauvin I have a bit of time this afternoon if you want to try figuring this out over slack. Up to you.

Issue #178, opened 2022-04-21
Explicit parallel interface


huard:
* RavenPy version:
* Python version:
* Operating System:

### Description

Instead of guessing which parameters are parallel, I suggest we ask the user to explictly tell RavenPy which parameters should be parallelized. 

For example, the `run` method could have a parallel argument holding a dictionary of `{key: values}` where keys are parameter names, and values list of parameters to iterate over. 

So 
`model.config.rvt.nc_index = [1,2,3]` would indicate RavenPy to use all three indices when creating Gauge commands, while
`model(ts=TS, parallel={'nc_index': [1,2,3]})` would tell RavenPy to run the emulator three times with `nc_index` set to 1, 2, 3 respectively in each simulation.

PR #180, opened 2022-04-21
Create multiple gauges when netCDF includes multiple stations.

### LINKED ISSUE #177: Allow multiple gauge stations
* RavenPy version:
* Python version:
* Operating System:

### Description

Raven can work with multiple gauge stations, using the `:Interpolation` command to weight the different stations. By default, Raven uses `INTERP_NEAREST_NEIGHBOR` to select the station closest to each subbasin. 

This is not supported at the moment by RavenPy. To support it, one option is to repurpose the `nc_index` config parameter to designate the index or the list of indices used to create `Gauge` commands. 

A side effect of this is that `nc_index` could not be used anymore to run parallel simulations with multiple input time series. This functionality is used in ensemble forecasting, hindcasting and data assimilation.

This underlines the fact that parallelism in RavenPy is inferred from the model inputs. For example, if `params` is a list of Params objects, then RavenPy assumes that it should run multiple times with each set of parameters. If `nc_index` is a list, then RavenPy assumes it should run in parallel with each value in the list. I think the parallel interface should be  more explicit.

huard:
Default is still to use first station only, but we could change it to use all stations by default. 

With this PR, passing `nc_index=[0,1,2]` will create three `:Gauges`, one for each index.  

Depends on #178
Fixes #177


huard:
@Bijan55699 

Will you be able to add the famine input dataset to raven-testdata when it's ready? Then we can make sure this works as expected. I've already tested with a local version.

Bijan55699:
> @Bijan55699
> 
> Will you be able to add the famine input dataset to raven-testdata when it's ready? Then we can make sure this works as expected. I've already tested with a local version.

@huard : Sure. I'll add it to the raven-testdata.

cjauvin:
@huard @Bijan55699 Should a `famine/famine_input.nc` be present in raven-testdata? https://github.com/Ouranosinc/raven-testdata/tree/master/famine Doesn't seem to be there. I'm merging this anyway.

Issue #181, opened 2022-05-02
Calibration using spotpy sees nonlinear slowdown as with OSTRICH


richardarsenault:
I have added a test "test_spotpy_calibration.py" that performs the calibration of a Raven model using DDS. 

The code works well and returns good parameter values. It could probably be improved (make the objective function more flexible) but for now the proof of concept is there.

However, even running locally, I see a major slowdown depending on the number of model evaluations. The increase in time is non-linear. Here are some test results:

10 evals       = 4.6 seconds        -->0.46 sec/eval
100 evals     = 49 seconds         -->0.49 sec/eval
1000 evals   = 763 seconds       -->0.76 sec/eval
1200 evals   = 1000 seconds     -->0.83 sec/eval
2500 evals   = 2951 seconds     -->1.18 sec/eval
5000 evals   = 9580 seconds     -->1.92 sec/eval
7500 evals   = 20004 seconds   -->2.66 sec/eval
10000 evals = 35000 seconds   -->3.50 sec/eval

As can be seen from my tests, initially the model takes about 0.46s per evaluation, but at the end, sims take (on average) 3.5 seconds each. Since this is the overall mean, the actual time taken by the final runs is likely much longer.

These times also only include the actual calibration time, and not any of the data processing or hyperparameter adjustment.

The code is in the branch "add_spotpy" and includes both the test file and the spotpy calibration logic as well in the ravenpy utilities.

Since this slowdown also occurs on the server, I am wondering if it could be caused by the way we manage the symlinks and file copies etc.? Perhaps there is a point where results are piling up on each other and trying to append to a progressively larger list, things like that? I am out of my element here but I don't see any other options.

Hopefully this helps investigating and solving the slowdown issue.

richardarsenault:
@huard I assigned you just so you see this, feel free to redirect!

PR #182, opened 2022-05-02
Do not write rv zip file and merge outputs if suppress_output is True

huard:
Seems to solve the non-linear slow-down described in #181


huard:
Shoud resolve https://github.com/Ouranosinc/raven/issues/168

huard:
Do we want to create the zip file from scratch, instead of append to it ?

richardarsenault:
I am not sure! In some cases it might be useful to have a log of what all results are, in which case we set suppress_output to false. For most cases it will be a single run, and for most calibrations we won't need them at all. So I think the proposed solution here is the best of all worlds.

richardarsenault (approved):
Works perfectly, suggest we merge this so we can merge spotpy and then upgrade notebooks at a later time.

huard:
Sorry, didn't get notified of your comment. Will do. 
Everyone is gone, so I'll take care of the nb next week.

richardarsenault:
@huard Hmmm one thing we could do (and that I do in my notebooks) is that at the end of the calibration, re-do a simulation with all the same data and optimal parameters but with suppress_output = False. Could we simply hardcode that, and return the diagnostics along with the original outputs of the calibration method?

huard:
That would work, but from a user point of view, you would set suppress_output to True, then after calling the model, you'd end up with suppress_output being False. I think this solution is ok as long as it is contained with the WPS service, but I would not do that in RavenPy.

PR #183, opened 2022-05-02
Add the SPOTPY calibration and sensitivity package and fix the calibration non-linear time increase

richardarsenault:
This PR solves #181 and #182. It adds the SPOTPY calibration and sensitivity analysis package, which is lightweight and easier to work with than OSTRICH. It also has a list of calibration algorithms such as DDS and SCEUA. It also does sensitivity analysis and seems quite easy to adjust to new functions and algos. 

The code was at first affected with the same non-linear slowdown during long calibration runs, but this PR also adds a fix to that problem by removing some processing when config.rvi.suppress_outputs = True, which scaled non-linearly. 

Now calibrations are faster, easily adaptable, do not require specific templating as in Ostrich, and scale linearly with the number of model evaluations.


huard:
Did a bit of clean-up.

Zeitsperre:
@huard @richardarsenault  is this safe to merge?

PR #184, opened 2022-05-04
modernize coveralls reporting, remove travis.yml config

Zeitsperre:
The coveralls GitHub action for reporting coverage has inconsistent behaviour. I figured it would be a good idea to port a working non-action approach here.


Zeitsperre:
@huard this is good for review. Coverage reporting is now fixed.

huard (approved):

PR #193, opened 2022-08-04
Unpin Numpy and NetCDF4 dependencies

Zeitsperre:


cjauvin:
So should we merge this? Is there a reason why it's still a draft?

Zeitsperre:
I haven't touched the numpy part yet; I simply unpinned numpy in the environment.yml. We still need to deal with the `# FIXME` I added.

Zeitsperre:
So for some reason the installation process for RavenPy force-installs an older numpy???

https://github.com/CSHS-CWRA/RavenPy/runs/7699123779?check_suite_focus=true#step:6:4064

cjauvin (approved):

PR #196, opened 2022-08-17
Redirecttofile

huard:
Implements `:RedirectToFile` command for grid_weights.


huard:
@Zeitsperre Any idea why the installation would fail ?

huard:
Failures are due to network issues on our end.

richardarsenault (approved):

huard:
So the DAP test failures were not entirely due to network issue, but also due to the fact that input files were converted to Path instances, and only their name written to the rvt. This works for files symlinked in the directory, but not for urls. Fixed in a85c08e6ad15d2231d4081e5d30286ac9a0e854e

PR #199, opened 2022-09-16
soil profile (Aida's work) and evaluation metrics

Bijan55699:
Addition of two new soil profiles (Aida's work) and activation of evaluation period command.

### DIFF
--- a/ravenpy/config/rvs.py
+++ b/ravenpy/config/rvs.py
@@ -250,7 +250,7 @@ class RVI(RV):
     # come after `:SoilModel`
     _post_tmpl = """
     :EvaluationMetrics     {evaluation_metrics}
-    {evaluation_periods}
+    :EvaluationPeriod      {evaluation_periods}
     :WriteNetcdfFormat     yes
     #:WriteForcingFunctions
     :SilentMode
--- a/ravenpy/models/emulators/hbvecmod.py
+++ b/ravenpy/models/emulators/hbvecmod.py
@@ -196,6 +196,16 @@ class HBVECMOD(Raven):
                 ["TOPSOIL", "FAST_RES", "SLOW_RES"],
                 [P.X03, P.X04, P.X05],
             ),
+            SOIL(
+                "loam",
+                ["TOPSOIL", "FAST_RES", "SLOW_RES"],
+                [P.X03, P.X04, P.X05],
+            ),
+            SOIL(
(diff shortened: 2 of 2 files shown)


Bijan55699:
soil profile (Aida's work) and evaluation metrics

ravenpy/config/rvs.py
```diff
@@ -250,7 +250,7 @@ class RVI(RV):
     # come after `:SoilModel`
     _post_tmpl = """
     :EvaluationMetrics     {evaluation_metrics}
-    {evaluation_periods}
+    :EvaluationPeriod      {evaluation_periods}
```
huard, line 253: Unless there is a problem with the original formulation, this should not be changed. 

This is the definition of the EvaluationPeriod command:
 ```
@dataclass
class EvaluationPeriod(RavenCommand):
    """:EvaluationPeriod [period_name] [start yyyy-mm-dd] [end yyyy-mm-dd]"""

    name: str
    start: dt.date
    end: dt.date

    template = ":EvaluationPeriod {name} {start} {end}"

    def to_rv(self):
        return self.template.format(**asdict(self))
```
In practice, use something like (untested):
```
model.config.rvt.evaluation_periods = [EvaluationPeriod('wet', dt,date(1990, 1, 1), dt.date(2000, 1, 1))]
```

huard (changes requested):

huard:
@Zeitsperre Seems to be a problem with the netcdf build. No objection pulling 3.7 from the suite if that simplifies things.

Zeitsperre:
I believe it's this issue here: https://github.com/Unidata/netcdf4-python/issues/1179. The wheel is bad on PyPI. I can add the fix for it in here at some point.

Zeitsperre:
@huard Python3.7 is rapidly approaching [end of life](https://peps.python.org/pep-0537/), so let's do that and see if the issue stems from there. I'm going to drop it in another PR.

ravenpy/config/rvs.py
huard: What do you mean by "original formulation" ? 
https://github.com/CSHS-CWRA/RavenPy/blob/427f7caf0562928db1951e42b1217727bf925c62/ravenpy/config/commands.py#L80

If there's a but with the EvaluationPeriod configuration, let's fix it in another PR.

PR #203, opened 2022-09-30
Address failing documentation builds

Zeitsperre:


Zeitsperre:
@huard This is good to go.

RtD builds are not being reported as status checks here because of a token-related issue. In order to fix that, the admins of https://github.com/CSHS-CWRA need to flush the currently active ones (or provide you and/or me with some additional access).

Until then, we just need to keep an eye on the RtD Build badge.

huard:
Thanks, 

Could you provide @boshek instructions regarding what needs to be done ?

Makefile
```diff
@@ -67,7 +67,9 @@ coverage: ## check code coverage quickly with the default Python
 docs: ## generate Sphinx HTML documentation, including API docs
 	# Warning: as the sphinx-apidoc is NOT being run on the RTD server the workaround we have
 	# is to commit the (currently 8) rst files it generates.
-	sphinx-apidoc -d 10 -o docs/ ravenpy
+	rm -f docs/ravenpy*.rst
```
huard, line 70: That looks dangerous, no ?

Zeitsperre: We use the exact line in all other docs-making entries (e.g. https://github.com/Ouranosinc/xclim/blob/b0ed02f3d8bc25667b0f7d87cfe294177ce76269/Makefile#L87). I could remove the `-f` for safety (this only affects files made when `sphinx-apidoc` is run, typically by the user who last called it). What do you think?

Zeitsperre: Actually, I'd keep the "-f" as some shells (`zsh`/`fish`) can potentially prompt for user input when removing multiple files. There's no recursion happening here. Should I instead make a mention in the docs to not name any doc files after the slug name (`ravenpy`)?

Zeitsperre: Sure, that seems reasonable.

huard (approved):

Zeitsperre:
> Could you provide @boshek instructions regarding what needs to be done ?

The ReadTheDocs OAuth app needs to be permitted to run at the organization level (https://docs.readthedocs.io/en/latest/integrations.html#webhook-activation-failed-make-sure-you-have-the-necessary-permissions). I'm not sure exactly how that is done, as we don't do that at Ouranosinc (our permissions are higher). If the admins themselves are not on ReadTheDocs, they may need to join it to allow access to the organization.

Seeing as David and I aren't members of CSHS-WRA, we can't make App Authorization/Integration requests at the organization level, so our ability to manage this is limited.

PR #206, opened 2022-10-24
undo suppress_output changes from 0.8.0, as it breaks raven-wps

huard:
Spotpy calibration repeatedly calls model(), and each call, rv files are created and added to a zip file. As the zip file grows, adding new files slows down. In 0.8.0, I fixed this by skipping `merge_output` if `suppress_output` is True. This has the unintended side effect of breaking multiple tests in raven-wps and would be backward incompatible for raven-wps users, at least, without fixes to the wps handlers. 

The solution proposed here is to create another method `Raven._execute`, which runs the model but does not parse the outputs. The Spotpy calibration can use this method and avoid the slowdown problem. I've modified diagnostics so it parses output files on demand.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 History
 =======
 
+0.8.1
+-----
+
+* Undo change related to `suppress_output`, as it breaks multiple tests in raven. New `Raven._execute` method runs models but does not parse results.
+
+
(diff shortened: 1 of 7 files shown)


richardarsenault (approved):
Elegant fix, Looks good to me!

PR #209, opened 2022-10-25
Automatically infer scale and offset from units attribute

huard:
When reading a netCDF CF-compliant input file, parse the units and infer scale and offset parameters. 

Todo: 
- [x] Add change to history
- [x] Remove explicit scale and offset configuration from tests
- [x] Merge https://github.com/Ouranosinc/raven-testdata/pull/28


Zeitsperre:
@huard You may want to pin `pint` in this PR. See: https://github.com/hgrecco/pint/issues/1631

Changes will be available in the newest xclim once https://github.com/Ouranosinc/xclim/issues/1211 is resolved.

PR #217, opened 2022-10-28
Refactor HRUState so that it prints only relevant state variables

huard:
**This is a breaking change.** 

Change the signature of HRUState. 
Instead of `HRUState(soil0=10)`, we now write `HRUState(data={"SOIL[0]": 10})`. 

Knowing which variables are specified, we're able to only put those in the rvc. This avoids warnings of `initial conditions for state variables not in model`. I think it's also a better design become the variable names are identical to those in the RavenC documentation.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,11 +4,24 @@ History
 
 0.9.0 (unreleased)
 ------------------
+
+Breaking changes:
+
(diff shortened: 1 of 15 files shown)


huard:
@ldajab This will avoid having to redefine the RVC template.

richardarsenault (approved):

PR #223, opened 2022-11-03
Remove RAVENPY_TESTDATA_PATH and rewrite tests for thread safety

Zeitsperre:
## Changes

 * Testing ensemble no longer requires that a `RAVENPY_TESTDATA_PATH` be set, nor does it require a local git clone of the raven-testdata
 * a threadsafe temporary folder for housing data is now implemented among the pytest fixtures
 * `pytest-xdist` is now a development dependency (allows for distributed testing). Improvements should already be visible on GitHub Actions (2 CPUs per VM).

## Potential issues 

For tests that used `get_local_testdata` with wildcards to access datasets, these data access steps were replaced with `query_folder` with regex patterns. Without authentication, this function is rate-limited, and it's possible that the rate may be too low for all of our synonymous builds. If that's the case, authenticating these requests for the GitHub API is relatively simple (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting), and I can look into that in another PR.

### DIFF
--- a/ravenpy/utilities/data_assimilation.py
+++ b/ravenpy/utilities/data_assimilation.py
@@ -7,13 +7,13 @@ import datetime as dt
 import math
 import os
 from copy import deepcopy
-from dataclasses import replace
 from typing import Dict, List, Sequence, Tuple, Union
 
 import numpy as np
 import xarray as xr
 
 from ravenpy.config.commands import HRUState
+from ravenpy.models import Raven
 
 """
 model = Raven model instance, preset with parameters etc.
@@ -109,9 +109,7 @@ def assimilate(
     return [xa, model]
 
 
-def perturbation(
-    da: xr.DataArray, dist: str, std: float, seed: int = None, **kwargs: Dict
-):
+def perturbation(da: xr.DataArray, dist: str, std: float, seed: int = None, **kwargs):
     """Return perturbed time series.
 
     Parameters
@@ -125,7 +123,7 @@ def perturbation(
     seed : int
       Seed for the random number generator. Setting the same seed for different variables will ensure the same
       perturbations are generated.
-    kwargs : dict
+    **kwargs
       Name and size of additional dimensions, apart from time.
 
     """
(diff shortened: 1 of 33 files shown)


Zeitsperre:
@huard All that needs to happen is merging of https://github.com/Ouranosinc/raven-testdata/pull/30. Testing time has been reduced by ~3.5 minutes!

huard (approved):
Could you include some info on what the changes is exactly. For example, if I run locally the same tests over and over, are test files cached on disk in between those runs? If a test file changes upstream, is it automatically detected ?

Zeitsperre:
@huard 

I was thinking about this last night, and I'm pretty sure I can do better on this PR. Right now, each running worker fetches their own copy of the dataset directly, but we could have a primary worker fetch all the relevant datasets on pytest launch to cache them on disk more permananently, then have each worker copy that to their temp directory when running. There would be no potential issues from `query_folder` API rate limiting, as it's called a handful of times once.

The best method I can think would be to create few pytest fixtures and/or load filenames into the pytest namespace. This will change how we access data, but will remove *a lot* of duplicated code. 

In xclim's `_get`, I added some changes that attempted re-download in the event that a md5 file fetched remotely does not match the signature for a file on disk, so I'll port those advancements here.

I'll re-request a review when my changes are all in.

Zeitsperre:
@huard This PR is about done. I can't firgure out what the problem is with the last failing tests. Could it be that the wrong files are being used to initialize the model? I can't seem to figure it out.

Zeitsperre:
There remains one final bug to deal with:

 * The raven RV files mostly use the following naming convention, except for those under `raven-hbv-ec`:
```
f"raven-{model}/raven-{model.split("-")[0]}-salmon.rv*"
```
If we can rename those files to follow the same pattern, the testing ensemble will pass, and that would be swell, IMO.

Zeitsperre:
@huard This PR is ready! It requires my PR in raven-testdata (https://github.com/Ouranosinc/raven-testdata/pull/32) to be merged. Will update the history and other documentation and merge when ready.

Zeitsperre:
Fixed. I think there are two issues:
 - Climpred is not compatible with the newest xarray (2022.11.0)
 - NetCDF4 >= 1.6 has major issues (PyPI wheel is definitely broken)

I'll try a few more variations, but we can at least say with certainty that the issue isn't on our end.

PR #226, opened 2022-11-04
Import changes to routing product extractor from aida-famine branch

huard:
Another step in bringing the master closer to the famine branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,8 +78,16 @@ jobs:
           environment-file: environment.yml
       - name: Conda and mamba versions
         run: |
-            mamba --version
-      - name: Install and test RavenPy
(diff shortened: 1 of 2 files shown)


Bijan55699:
@huard : I tried to test the modifications implemented in this PR, but the `hbvecmod.py `emulator does not exist in `adjust_routing_importer` branch. I think we will have to merge the latest changes of the `hbvecmod.py` in `aida-famine` branch so that I can add it to the `adjust_routing_importer `for testing (I added the the latest file after git pull, but looks like its not its missing two soil types and potentially other changes made)

huard:
This PR is a copy paste of the file in the aida-famine branch. I just wanted to confirm with you that it was ok to bring those changes back into master.

PR #231, opened 2022-11-08
Support distinct netcdf indices for meteo and hydrometric stations

### LINKED ISSUE #228: Support observation datasets with multiple gauge stations
### Description

At the moment, there is only one index into station data (`nc_index`). If we have a file for meteo data and another for streamflow data, we need another index to pinpoint the gauge stations. 

I propose creating a second index (`hydro_idx`) to identify the gauge stations to be used as observations.

huard:
Closes #228

- Allows configuration to use distinct files and indices for hydro and meteo data. 
- Adds support for multiple hydro gauge stations. To use, configuration needs to include a new config parameter called `gauged_sb_ids`, a sequence identifying the subbasin draining at the gauge station.
- Renames configuration parameter `nc_index` to `meteo_idx` to reduce confusion between the two indices. `nc_index` should still work but is considered deprecated and will be removed in a future version.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,7 +12,9 @@ Breaking changes:
   dictionary keyed by variables' Raven name (e.g. `{"SOIL[0]": 10}).
   This change makes `rvc` files easier to read, and avoids Raven
   warnings regarding 'initial conditions for state variables not in model'.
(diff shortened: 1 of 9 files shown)


richardarsenault (approved):
This is a great idea, thanks!

PR #233, opened 2022-11-14
Prepare version 0.9.0

Zeitsperre:
## Changes

* Notebooks have been re-run and adapted to reflect API changes
* Documentation building is now more step-based runs a linkcheck on build
* Pydantic classes are parsed by Sphinx using `autodoc_pydantic` to prevent warnings on ReadTheDocs
* Docstrings have been adjusted to better follow conventions
* Bumped version to 0.9.0


huard:
Thanks for updating the nb ! There are a bunch of new warnings in some of them that I should investigate before the release. Let's hold on for a bit.

Zeitsperre:
All good! I want to get the docs working again, so still plenty of things to do on my end as well. If you want to open a PR basing your changes off this branch, feel free!

Zeitsperre:
The documentation is now working! Pydantic classes were being double-documented by Sphinx and printing warnings because of this. We need to ensure that Pydantic classes never employ numpy-style docstrings (their methods can use numpy-docstring though).

huard:
Well done. The "no docstring in pydantic classes" seems like a bug, however. Is this a known failure or its something you discovered?

Zeitsperre:
It's not that there wasn't any docstring, it was that the numpy docstring was conflicting with the docstring that was being generated by Pydantic via type declaration. It's a known issue for Sphinx (https://github.com/mansenfranzen/autodoc_pydantic). The build errors were being raised since there were two targets being generated for the same classes (those using Pydantic).

In the meantime, I'll move on to other tasks from here while you deal with the notebook warnings.

huard:
All good, I can work with their documentation standard. This is really good to know in advance because I'm working on pydantic models for every Raven command.

PR #256, opened 2023-01-30
Update Fiona to v1.9

Zeitsperre:
Fiona v1.9 has API changes that necessitate changes on our end. This will render us more in line with Fiona v2.0 when it is soon released. The primary change is that Fiona features are now their own class `fiona.Feature` containing `fiona.Geometry` classes. These are compatible with `dict()` outputs, so no changes should be needed in `Raven-WPS`.

For more information, see: https://github.com/Toblerity/Fiona/pull/787

### Changes

- Set minimum GDAL version at 3.1
- `pip` was comparing compatible versions of `sphinx_rtd_theme`, so this has now been set a `>=1.0`.
- The new Fiona API is now used for vector reprojection. `Fiona.write` is now used for writing output file JSONs.

### Additonal Notes
~The conda package is not yet released due to a few small changes needed in their testing suite. This should be addressed soon. See: https://github.com/conda-forge/fiona-feedstock/pull/208~ - This has been released on conda-forge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,10 +68,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Setup miniconda with mamba
-        uses: conda-incubator/setup-miniconda@v2.1.1
+        uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: "3.9"
-          mamba-version: "*"
+          miniforge-variant: Mambaforge
           channels: conda-forge,defaults
           channel-priority: true
           activate-environment: ravenpy-env
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   black:
     name: Code linting
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Cancel previous runs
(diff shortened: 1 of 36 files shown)


Zeitsperre:
@huard 

The present failures are due to service issues with our THREDDS server. Should I make these tests `XFAIL`able?

huard:
The error I'm seeing has to do with cf-xarray:
```No such file or directory: '/home/runner/work/RavenPy/RavenPy/.tox/py38/lib/python3.8/site-packages/cf_xarray/data/cf-standard-name-table.xml'```

Zeitsperre:
The THREDDS issues have been resolved, cf-xarray fixes are on the way: https://github.com/xarray-contrib/cf-xarray/issues/415

I'm wondering though if we should allow the THREDDS-related tests to fail since they're service-dependent. What do you think?

huard:
Is there a way to fail only if the service is not on ? If the service is available, but the test fails, I'd like to catch that.

Zeitsperre:
I can definitely look into that. If it's simple enough that I can add it here this morning, I'll add it. Otherwise, no more changes. As it stands, though, am I good to merge?

huard (approved):

Zeitsperre:
Failing because of a new `black`. Opening another PR to fix this...

PR #264, opened 2023-02-15
Replace setup.py with pyproject.toml using scikit-build-core (experimental)

Zeitsperre:
### Changes

* `setup.py` and the `setuptools`-based installation and development scripts for RavenC and Ostrich
* Add a CMake and Ninja based method for fetching the RavenC source files and building them against NetCDF on `pip install`
* Add a BSD-Licensed CMake script for finding NetCDF on multiple platforms
* Update pre-commit and GitHub Workflows to reflect these changes
* Bump version to 0.12.0

### What's missing?

- [ ] `Ostrich`-based functionality must be removed from the library.
  - Ostrich is soon to be deprecated and has been removed from the installation process entirely. Tests that depend on it will absolutely fail at present.
- [ ] `--editable` installations must be supported. 
  - `scikit-build-core` does not currently support PEP660 (see: https://github.com/scikit-build/scikit-build-core/issues/171) but is planned to be supported soon.

### Why?

With new standards for tooling and packaging, the `setuptools`-based compilation approach was very quickly becoming unmaintainable and only supported building on Linux/Unix platforms. `scikit-build-core` is a setup backend that supports multi-OS builds using CMake (which is a much more approachable multiplatform C/C++ build engine). Assuming that the needs for this implementation are eventually met, this will render the `raven-hydro` conda-forge package obsolete, as we would be able to offer Python wheels that package `RavenC` or source distributions that allow users to compile it themselves on `pip install` (similar to `numpy`).


huard:
Made obsolete by https://github.com/Ouranosinc/raven-hydro

PR #267, opened 2023-03-06
Fix file-access errors with newest NetCDF4, adjust testing default settings

Zeitsperre:


huard (approved):

Zeitsperre:
I've been running into an issue like this one: https://github.com/Unidata/netcdf4-python/issues/1205. Hopefully, the workaround of ensuring that files are closed properly will address this?

Zeitsperre:
@huard

I decided to throw in the testing adjustments so that individual tests do not create multiple workers. Similar to changes in https://github.com/Ouranosinc/xclim/commit/b69b86744ce643d8ca0768a966825df28feb9c78, we now have `pytest-xdist` disabled when running `pytest` and enabled when running `tox`.

I also modified the env vars:
 - `MAIN_TESTDATA_BRANCH` --> `RAVENPY_TESTDATA_BRANCH`
 - `SKIP_TEST_DATA` --> `RAVENPY_SKIP_TEST_DATA`

These env vars can be set in the main CI workflow.

PR #278, opened 2023-04-19
Adopt PEP 517 and PEP 621, use raven-hydro package

### LINKED ISSUE #283: Support Raven 3.7
* RavenPy version:
* Python version:
* Operating System:

### Description

Describe what you were trying to get done.
Tell us what happened, what went wrong, and what you expected to happen.

- Pin RavenPy to Raven 3.7;
- Add support for new options; 

### What I Did

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

Zeitsperre:
Closes #283 

## Changes

* Replaced `setup.py` with new format `pyproject.toml` using the `flit` backend.
* Updated pre-commit hooks to include formatters and checkers for TOML files.
* Removed the obsolete `requirements*.txt` and `Manifest.in` files.
* Dealt with an import-based error that occurred due to the sequence in which modules are loaded at import (attempting to call `ravenpy` before it is installed).
* The build recipes no longer build on each other, so when installing the `dev` or `docs` recipe, you must also install the `gis` recipe. A workaround I've added is to simply list the `gis` recipes in both for now
* This PR also updates the GeoServer API calls to at the very least work with the newest GeoPandas (using `urllib`). A more elegant solution should be looked into (to synchronize `owslib / requests / urllib.request` calls).
* This PR leverages the new method of installing the Raven model using https://github.com/Ouranosinc/raven-hydro.

## What still needs to be addressed

* ~There is now no method of building `raven-hydro` from PyPI. The current approach was rendered obsolete in `pip>=23.1`, so we are in the right to retire this method.~
  * A new method of installing `raven-hydro` using `scikit-build-core` is in the works.
    * This has been implemented (https://github.com/Ouranosinc/raven-hydro); A release on PyPI is forthcoming.
* ~The CI for the PyPI builds using `tox` are broken due to the missing `raven-hydro` binary. There is likely a whole slew of things that must be addressed in `tox` (removal of tons of now-obsolete workarounds), so don't expect this to work yet.~
  * These issues have been resolved with the new installation method (https://github.com/Ouranosinc/raven-hydro); A release on PyPI is forthcoming.
* ~There are documentation files as well as the top-level `Makefile` that refer to operations using `setup.py`, and `twine`. These need to be replaced with their `flit`-based equivalents. For more information, see: https://flit.pypa.io/en/stable/index.html~
  * These issues have been resolved.

## Why?

With new standards for tooling and packaging, the setuptools-based compilation approach was very quickly becoming unmaintainable and only supported building on Linux/Unix platforms.

For [raven-hydro](https://github.com/Ouranosinc/raven-hydro), scikit-build-core is a setup backend that supports multi-OS builds using CMake (which is a much more approachable crossplatform C/C++ build engine). This renders the setuptools-based installation approach obsolete, as we are now able to install a Python wheel or source distribution specifically for the Raven model that allow users to compile it themselves on pip install (similar to numpy).


Zeitsperre:
The new `fiona`-related errors are due to changes in `geopandas` that affect reading files from URLs: https://github.com/geopandas/geopandas/pull/2796

The fix is to simply adjust the way we handle these objects in `ravenpy.utilities.geoserver`. Benefits include significant read speedups.

pyproject.toml
```diff
@@ -0,0 +1,226 @@
...
+  "pandas<2.0",
+  "pint>=0.20",
+  "pydantic",
+  "pymbolic",
+  "requests",
+  "scipy",
+  "spotpy",
+  "statsmodels",
+  "wheel",
+  "xarray<2022.11.0", # Pinned due to incompatibility with climpred @ 2.2.0
```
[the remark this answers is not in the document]
huard: The latest version of climpred pins array to <2023.01, so we may not have to pin it ourselves.

tests/conftest.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 import xarray as xr
 from filelock import FileLock
-from xclim.indicators.land import fit, stats
+from xclim.indicators.generic import fit, stats
```
huard, line 10: Is this because there is deprecation warning?

docs/installation.rst
```diff
@@ -33,9 +33,9 @@ Custom Installation (Python/Pip)
 --------------------------------
 
 .. warning::
-   The following instructions will only work on POSIX-like systems (Unix/Linux; not supported on Windows).
+   As of April 2023, this installation method does not install the `raven-hydro` hydrological model. You must install the `raven-hydro` model and add it to your `bin` or append it to your `$PATH` manually. This may change in the future
 
-If you wish to install RavenPy and its C-libraries manually, compiling the `Raven` binaries for your system,
+If you wish to install RavenPy and its C-libraries manually, compiling the `Raven` binary for your system,
 you can install the entire system directly, placing them in the `bin` folder of your environment.
 In order to perform this from Ubuntu/Debian:
```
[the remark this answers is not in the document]
huard: Just make sure that the Install docs are clear on what needs to be done now, ie. use conda or compile yourself.

huard (approved):

tests/conftest.py
Zeitsperre: This is because we have just changed it: https://github.com/Ouranosinc/xclim/commit/0e06e2c54fc97fe0b5c67fd2802ca342622ff337

I think we need to pin this to the newest version of xclim as a result!

Zeitsperre: Either that or we can make it backwards-compatible with a try/except block.

huard: Ok, let's pin to the newest xclim then.

Zeitsperre:
I think I've earned this.

its_working_star_wars

Zeitsperre:
Excepting changes to expected values that will be in the new Raven-Hydro version and a few platform-specific bugs, everything is working!

Zeitsperre:
Pydantic failures on conda have already been addressed. Conda package soon: https://github.com/pydantic/pydantic/issues/5824

Zeitsperre:
There's an environment inconsistency for conda:
- The newest `raven-hydro` is built against the newest libnetcdf/libhdf5/etc.
- The last build of `climpred` on conda-forge was in November 2022 (no updates since)

As it stands, the libraries are incompatible:
```
  - climpred               2.1.6  pyhd8ed1ab_1                 conda-forge                    
  + climpred               2.3.0  pyhd8ed1ab_0                 conda-forge/noarch       Cached
  - raven-hydro            0.2.0  py310h3ce4ad4_0              conda-forge                    
  + raven-hydro              3.6  ha2b8ddb_1                   conda-forge/linux-64     Cached
```

I think climpred needs to be regenerated, or we need to offer pinned versions of netCDF (which I really very much don't want to do).

Zeitsperre:
@huard

It's looking good! The last thing that needs to be revised is the one failing test on macOS. I think it has to do with the CRS not being read on file open. One approach would be to change the engine being used to read the file (try GeoPandas?) Will explore this tomorrow.

huard (approved):

PR #298, opened 2023-07-19
Skip manual install of NetCDF-python binaries on macos

Zeitsperre:


huard (approved):

Zeitsperre:
I'm going to merge this and open an issue. It looks like this is a bug that is affecting OPeNDAP, likely from netcdf4-python.

PR #306, opened 2023-09-07
Added the paper pdf to the notebook

richardarsenault:
Added the paper for users to reference from the platform itself.


Zeitsperre:
Hey @richardarsenault,

I'm waiting on @huard to fix a test or two before a few PRs can go forward; Those PRs also fix the pre-commit hook (it shouldn't be modifying the notebooks in this PR).

Would you be willing to add a link somewhere in the documentation to this file? See: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-download

richardarsenault:
@Zeitsperre I could add a link somewhere, but for this ideally it would need to be hosted? So perhaps I should wait for the PRs to finish such that this file can have a permanent spot to link to?

thanks!

Zeitsperre:
Closing in favour of #311

PR #312, opened 2023-09-22
Configurable servers

Zeitsperre:
FYI @fmigneault

### Changes
* The THREDDS and GeoServer instances can now be configured for staging/testing purposes. These new environment variables are documented within the installation docs:
    * `RAVENPY_THREDDS_URL`: URL to the THREDDS-hosted climate data service. Defaults to `https://pavics.ouranos.ca/twitcher/ows/proxy/thredds`.
    * `RAVENPY_GEOSERVER_URL`: URL to the GeoServer-hosted vector/raster data. Defaults to `https://pavics.ouranos.ca/geoserver`.
        * This environment variable `GEO_URL` is still valid for legacy purposes.
* The `_determine_upstream_ids` function under `ravenpy.utilities.geoserver` was a direct copy of `determine_upstream_ids` found in `ravenpy.utilities.geo` and has been removed.
* The `get_CASPAR_dataset` and `get_ECCC_dataset` functions now accept a `thredds` and a `directory` in their call signatures for specifying file location.
* Pins `xarray` below `2023.09.0` until xclim addresses compatibility issues (will be fixed in `0.46.0`).

### Relevant discussion:

https://github.com/bird-house/birdhouse-deploy/issues/14?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDY5OTA0MTYxODoxMDgxOTUyNA%3D%3D#issuecomment-1730183776

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,15 @@ History
 
 0.12.4 (unreleased)
 -------------------
-* In tests, set xclim' missing value option to ``skip``. As of xclim 0.45, missing value checks are applied to ``fit`` indicator, meaning that parameters will be set to None if missing values are found in the fitted time series. Wrap calls to ``fit`` with ``xclim.set_options(check_missing="skip")`` to reproduce the previous behavior of xclim.
+
+Breaking changes
+^^^^^^^^^^^^^^^^
+* In tests, set `xclim`'s missing value option to ``skip``. As of `xclim` v0.45, missing value checks are applied to the ``fit`` indicator, meaning that parameters will be set to `None` if missing values are found in the fitted time series. Wrap calls to ``fit`` with ``xclim.set_options(check_missing="skip")`` to reproduce the previous behavior of xclim.
+* `RavenPy` processes and tests that depend on remote THREDDS/GeoServer now allow for optional server URL and file location targets. These can be set with the following environment variables:
+    * `RAVENPY_THREDDS_URL`: URL to the THREDDS-hosted climate data service. Defaults to `https://pavics.ouranos.ca/twitcher/ows/proxy/thredds`.
+    * `RAVENPY_GEOSERVER_URL`: URL to the GeoServer-hosted vector/raster data. Defaults to `https://pavics.ouranos.ca/geoserver`.
+         * This environment variable was previously called `GEO_URL` and was renamed to narrow its scope to RavenPy.
+* The `_determine_upstream_ids` function under `ravenpy.utilities.geoserver` has been removed as it was a duplicate of `ravenpy.utilities.geo.determine_upstream_ids`. The latter function is now used in its place.
 
 0.12.3 (2023-08-25)
 -------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,11 +8,14 @@ History
 Breaking changes
 ^^^^^^^^^^^^^^^^
 * In tests, set `xclim`'s missing value option to ``skip``. As of `xclim` v0.45, missing value checks are applied to the ``fit`` indicator, meaning that parameters will be set to `None` if missing values are found in the fitted time series. Wrap calls to ``fit`` with ``xclim.set_options(check_missing="skip")`` to reproduce the previous behavior of xclim.
-* `RavenPy` processes and tests that depend on remote THREDDS/GeoServer now allow for optional server URL and file location targets. These can be set with the following environment variables:
-    * `RAVENPY_THREDDS_URL`: URL to the THREDDS-hosted climate data service. Defaults to `https://pavics.ouranos.ca/twitcher/ows/proxy/thredds`.
-    * `RAVENPY_GEOSERVER_URL`: URL to the GeoServer-hosted vector/raster data. Defaults to `https://pavics.ouranos.ca/geoserver`.
-         * This environment variable was previously called `GEO_URL` and was renamed to narrow its scope to RavenPy.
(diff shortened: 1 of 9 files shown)


ravenpy/extractors/forecasts.py
```diff
@@ -19,6 +21,12 @@
 
 LOGGER = logging.getLogger("PYWPS")
 
+# Do not remove the trailing / otherwise `urljoin` will remove the geoserver path.
+# Can be set at runtime with `$ env RAVENPY_THREDDS_URL=https://xx.yy.zz/geoserver/ ...`.
+THREDDS_URL = os.environ.get(
+    "RAVENPY_THREDDS_URL", "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/"
+)
```
fmigneault, line 28: The resolved `THREDDS_URL` should be checked and add the missing trailing `/` if needed to avoid hard to find errors if `RAVENPY_THREDDS_URL=https://xx.yy.zz/geoserver` (no trailing `/`) was used by mistake. The same comment applies anywhere `urljoin` is used with potential user-provided values.

fmigneault (commented):
Not sure how it relates to https://github.com/bird-house/birdhouse-deploy/issues/14 for the WPS response URLs?

ravenpy/extractors/forecasts.py
```diff
@@ -38,15 +46,36 @@ def get_hindcast_day(region_coll: fiona.Collection, date, climate_model="GEPS"):
...
+
+    Returns
+    -------
+    xr.Dataset
+        The forecast dataset.
+    """
 
     if climate_model == "GEPS":
         d = dt.datetime.strftime(date, "%Y%m%d")
-        file_url = f"https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/disk2/caspar/daily/GEPS_{d}.nc"
```
tlvu, line 49: Oh very nice removing the hardcode !!! This will allow for proper testing of new Thredds upgrade.

ravenpy/utilities/geoserver.py
```diff
@@ -6,7 +6,7 @@
 * BBox coordinates are passed as (lon1, lat1, lon2, lat2).
 * Shapes (polygons) are passed as shapely.geometry.shape parsable objects.
 * All functions that require a CRS have a CRS argument with a default set to WGS84.
-* GEO_URL points to the GeoServer instance hosting all files.
+* GEOSERVER_URL points to the GeoServer instance hosting all files.
```
tlvu, line 9: Same comment as the other https://github.com/Ouranosinc/raven/pull/485 about keeping backward-compat with the old public `GEO_URL`.

tlvu: This looks like an identical file as the other https://github.com/Ouranosinc/raven/pull/485, is this normal the same file is duplicated?

tlvu (changes requested):
Very nice now that both Thredds and Geoserver are configurable.  Simply keep the backward-compat with the old public config`GEO_URL` please.

ravenpy/extractors/forecasts.py
tlvu: Agreed with @fmigneault, defensive coding make the code more robust.

Zeitsperre:
@huard Builds appear to be failing due to SDBA-related changes. Could it be the latest `scipy` (10 days ago)?

HISTORY.rst
```diff
@@ -4,7 +4,15 @@ History
 
 0.12.4 (unreleased)
 -------------------
-* In tests, set xclim' missing value option to ``skip``. As of xclim 0.45, missing value checks are applied to ``fit`` indicator, meaning that parameters will be set to None if missing values are found in the fitted time series. Wrap calls to ``fit`` with ``xclim.set_options(check_missing="skip")`` to reproduce the previous behavior of xclim.
+
+Breaking changes
+^^^^^^^^^^^^^^^^
+* In tests, set `xclim`'s missing value option to ``skip``. As of `xclim` v0.45, missing value checks are applied to the ``fit`` indicator, meaning that parameters will be set to `None` if missing values are found in the fitted time series. Wrap calls to ``fit`` with ``xclim.set_options(check_missing="skip")`` to reproduce the previous behavior of xclim.
+* `RavenPy` processes and tests that depend on remote THREDDS/GeoServer now allow for optional server URL and file location targets. These can be set with the following environment variables:
+    * `RAVENPY_THREDDS_URL`: URL to the THREDDS-hosted climate data service. Defaults to `https://pavics.ouranos.ca/twitcher/ows/proxy/thredds`.
```
tlvu, line 12: This new `RAVENPY_THREDDS_URL` is brand new, why would it be breaking if there is nothing existing to break?

For the new `RAVENPY_GEOSERVER_URL`, since the old `GEO_URL` still supported, should normally not listed as a breaking change.

tlvu (approved):
LGTM.  We should always try to minimize breaking changes that are backward incompatible.

HISTORY.rst
tlvu: I saw the updated wording but I am not sure you understood my point.  My point was why `RAVENPY_THREDDS_URL` and `RAVENPY_GEOSERVER_URL` are listed in the "breaking change" section since they do not break any existing behavior.

For me "breaking change" means change of behavior that is not compatible with the previous behavior.  Not that "breaking change" means important changes that need to be highlighted.  Which definition of "breaking change" is in effect here?

tlvu (approved):
LGTM

PR #323, opened 2023-11-02
Handle scalar elevation values in nc_specs

### LINKED ISSUE #279: Having an elevation in the NC can cause issues
I extracted the data of a watershed from HYSETS_2020 and the resulting NC had an elevation of an xarray with 1 element. The method from_nc, crashes in the observationdata module, because it tries to get the value with [0] instead of item(0).

Removing the elevation from my NC and passing it in params, fixed the issue, but we shouldnt need to do this work around.

huard:
Fixes #279

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,14 +1,21 @@
 import pytest
+import xarray as xr
 
 from ravenpy.config.utils import nc_specs
 
 
-def test_nc_specs(get_local_testdata):
+def test_nc_specs(get_local_testdata, tmp_path):
     f = get_local_testdata(
         "raven-gr4j-cemaneige/Salmon-River-Near-Prince-George_meteo_daily.nc"
     )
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -1,16 +1,23 @@
 import pytest
+import xarray as xr
 
 from ravenpy.config.utils import nc_specs
 
 
-def test_nc_specs(get_local_testdata):
+def test_nc_specs(get_local_testdata, tmp_path):
```
huard, line 7:
```suggestion
def test_nc_specs(get_local_testdata):
```

Zeitsperre:
@huard It looks as though @Mayetea hasn't been on GitHub in some time. What do you want to do with this PR?

richardarsenault (approved):
This looks good, should definitely solve the problem with the elevation attribute. Other attributes from HYSETS would probably have the same problem BUT they are not needed to build a ravenpy model so I think this solution is perfectly fine.

PR #329, opened 2023-12-14
Fix notebook execution

Zeitsperre:
### Changes

 * Installs RavenPy using the ReadTheDocs `Python` block

### Discussion

Currently, the Jupyter (non-MyST) notebooks are not run (nbsphinx is not in the extensions. Is the plan to prgressively move these over to the MyST standard ?

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -17,8 +17,6 @@ build:
   tools:
     python: "mambaforge-4.10"
   jobs:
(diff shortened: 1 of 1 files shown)


huard:
Some of the nb are compute intensive, so not clear.

huard (approved):

huard:
Merci !

PR #337, opened 2024-01-12
Update dependencies

Zeitsperre:


Zeitsperre:
@huard 

For some reason, the Linux/Anaconda builds are segfaulting all over the place. I am completely at a loss. There are functionally no changes beyond the augmenting of a few versions (which are typically installed at a much higher version). There weren't any major software releases lately, so I can't say for certain it's even dependency-related.

My only theory is that it's affecting tests that depend on `python-netcdf4` and `geopandas` (both use `GDAL`). If this is another `GDAL` issue, I wouldn't be surprised.

Any other ideas?

huard (approved):

huard:
Thread issues with netCDF happening because xdist runs tests in parallel ? 

Also, why is the pydantic version not already set at > 2 ? I had done this in the pydantic PR already.

huard:
Ahh... I may have done the change in pyproject.toml but not elsewhere...

Zeitsperre:
> Thread issues with netCDF happening because xdist runs tests in parallel ?

macOS (conda) and the PyPI builds all run in parallel. No new releases of `libnetcdf` or `hdf5` in more than a month. I don't think it's due to that. Most of the errors are popping up when `GridWeightExtractor` is called. 

I can't test locally due to a weird bug with `libgxx-ng` (seeing this issue at `raven-hydro` as well). Something for next week.

> Also, why is the pydantic version not already set at > 2 ? I had done this in the pydantic PR already.

This must have flown under the radar. `pydantic` was likely being reinstalled by `pip` in our tests...

PR #338, opened 2024-02-15
Small fixes to support HBVECMOD emulator out of the box

huard:
- Add Interpolation command options. 
- Let VegetationClass records contain symbolic expressions


huard:
@lou-a for your review

PR #341, opened 2024-03-05
Install raven-hydro from official git sources

Zeitsperre:
### Changes

* Added support for Python 3.12 and dropped support for Python3.8.
* Upgraded `raven-hydro` to v0.3.0 and `RavenHydroFramework` to v3.8.
* `ravenpy` now requires `xclim` >= v0.48.2, `xarray` >= v2023.11.0, and `pandas` >= 2.2.0.
* Updated GitHub publishing workflows to use Trusted Publisher for TestPyPI/PyPI releases.
* Added Dependabot to keep dependencies up-to-date.
* Now using step-security/harden-runner action to harden GitHub Actions runners.
* Adjusted GitHub Workflows to test against Python 3.9, 3.10, 3.11, and 3.12.
* Updated the build-system requirements when testing with `tox` to use newer `setuptools` and `wheel` versions when building `gdal`.
* Adjusted the `test_emulators` expectations to handle the RavenHydroFramework version-dependent outputs.


pyproject.toml
```diff
@@ -37,52 +37,52 @@ dynamic = ["description", "version"]
...
+  "numpy",
+  "owslib >=0.24.1,<0.29", # see: https://github.com/geopython/OWSLib/issues/871
+  "pandas >=2.2.0",
+  "pint >=0.20",
   "platformdirs",
-  "pydantic>=2.0",
+  "pydantic >=2.0",
   "pymbolic",
-  "raven-hydro==0.2.4",
+  "raven-hydro ==0.2.4",
```
huard, line 54: Shouldn't it be 0.3.0 ?

huard (approved):

pyproject.toml
Zeitsperre: Precisely. It will be once we release it on PyPI and conda-forge.

PR #356, opened 2024-04-23
WPS returns datasets : access `band_data` variable

tlogan2000:
In notebook 02 using the PAVICS beta image `stats_resp.get(asobj=True)` returns an `xr.Dataset`  ... plotting requires accessing the `band_data` variable

See https://github.com/Ouranosinc/pavics-jupyter-env-issues/issues/7

Note this will break the notebook when using `current` image.  Do not merge until all fixes are ready

Fix for https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/121.


Zeitsperre:
@tlogan2000 The "current" image copies notebooks based on the `master` branch? Doesn't the RavenPy version lag by several months already?

We should change this setup.

tlogan2000:
> @tlogan2000 The "current" image copies notebooks based on the `master` branch? Doesn't the RavenPy version lag by several months already?
> 
> We should change this setup.

I'm not sure how this works exactly ... we'd need to ask @tlvu which version of the notebooks are pulled. Ideally the notebooks pulled would be based on a `ravenpy` release that is deployed in each pavics image but I have a feeling that it is straight from `master`

huard:
I suspect this might have to do with : https://github.com/bird-house/birdy/pull/240/files

huard:
This PR would revert to returning a DataArray
https://github.com/bird-house/birdy/pull/248

PR #358, opened 2024-04-23
Update dependencies, stabilize test

Zeitsperre:
## Changes

- Added some development dependencies that were missing to the `environment.yml`
- Upgraded `owslib` to `>=0.29.1`
- All operations that open NetCDF files or DAP links accept an `engine` argument. The default for all of these is `h5netcdf`
- Added `pydap` as an alternate backend for opening DAP links
- `test_climpred_hindcast_verif` is now skipped for Python3.10 builds. It seems to only fail on the particular version of Python. When examining the dependencies, other than the Python version (and ABI version), there are no differences in the environments between Python3.10 and Python3.11. My gut feeling points to an issue with `climpred`.
- Temporarily disabled tests for macOS on GitHub due to architecture changes.

### DIFF
--- a/ravenpy/config/commands.py
+++ b/ravenpy/config/commands.py
@@ -560,9 +560,31 @@ class ReadFromNetCDF(FlatCommand):
         cls, fn, data_type, station_idx=1, alt_names=(), engine="h5netcdf", **kwds
     ):
         """Instantiate class from netCDF dataset."""
-        specs = nc_specs(
-            fn, data_type, station_idx=station_idx, alt_names=alt_names, engine=engine
-        )
+        try:
+            specs = nc_specs(
+                fn,
+                data_type,
+                station_idx=station_idx,
+                alt_names=alt_names,
+                engine=engine,
+            )
+        except FileNotFoundError:
+            if validators.url(fn):
+                specs = nc_specs(
+                    fn,
+                    data_type,
+                    station_idx=station_idx,
+                    alt_names=alt_names,
+                    engine="pydap",
+                )
+            else:
+                specs = nc_specs(
+                    fn,
+                    data_type,
+                    station_idx=station_idx,
+                    alt_names=alt_names,
+                    engine="netcdf4",
+                )
         specs.update(kwds)
         attrs = filter_for(cls, specs)
         return cls(**attrs)
--- a/ravenpy/config/commands.py
+++ b/ravenpy/config/commands.py
@@ -827,8 +827,14 @@ class ObservationData(Data, coerce_numbers_to_str=True):
         """
 
     @classmethod
-    def from_nc(cls, fn, station_idx: int = 1, alt_names=(), **kwds):
-        specs = nc_specs(fn, "HYDROGRAPH", station_idx, alt_names)
+    def from_nc(cls, fn, station_idx: int = 1, alt_names=(), engine="h5netcdf", **kwds):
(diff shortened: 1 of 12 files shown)


ravenpy/config/commands.py
```diff
@@ -555,22 +555,26 @@ def reorder_time(cls, v):
...
+            alt_names=alt_names,
+            engine=engine,
+        )
         specs.update(kwds)
         attrs = filter_for(cls, specs)
         return cls(**attrs)
 
     @property
-    def da(self) -> xr.DataArray:
+    def da(self, engine="h5netcdf") -> xr.DataArray:
```
huard, line 574: You can't pass arguments to a property.

ravenpy/config/utils.py
```diff
@@ -48,6 +51,7 @@ def nc_specs(
 
     from ravenpy.utilities.coords import infer_scale_and_offset
 
+    i = 0
     # Convert to NumPy 0-based indexing
     if station_idx is not None:
         i = station_idx - 1
```
huard, line 57: Maybe instead move this lower inside the other "if station_idx is not None".

tests/test_hindcasting.py
```diff
@@ -80,14 +82,25 @@ def test_hindcasting_GEPS(self, get_local_testdata, salmon_hru, tmp_path):
...
     def test_climpred_hindcast_verif(self, get_local_testdata, salmon_hru, tmp_path):
         ts = get_local_testdata(
             "raven-gr4j-cemaneige/Salmon-River-Near-Prince-George_meteo_daily.nc"
         )
+        # Make a local copy to evade double-ownership of file - first file
+        ts_tmp1 = tmp_path / "salmon_river_near_prince_george-tmp1.nc"
+        ts_tmp1.write_bytes(ts.read_bytes())
+        # Make a local copy to evade double-ownership of file - second file
+        ts_tmp2 = tmp_path / "salmon_river_near_prince_george-tmp2.nc"
+        ts_tmp2.write_bytes(ts.read_bytes())
```
huard, line 103: Do you know why is this causing problems ? Should that copying logic be embedded in `get_local_testdata` directly ?

huard (approved):
Looks good. There's just the property argument that needs fixing.

tests/test_hindcasting.py
Zeitsperre: I added this in as an extra check, but I'm beginning to doubt whether this does anything. The issue appears to be that the file is modified in place and then re-opened later on _within the same function call__. Copying files passed to the function has no impact.

ravenpy/config/commands.py
Zeitsperre: Noted. Will remove.

tests/test_hindcasting.py
Zeitsperre: I'll leave it for now, as we still need to overhaul the system with `pooch`. I'll be removing lots of code like this then.

Zeitsperre:
@huard and @tlvu

GitHub has updated their macOS images to the `ARM64` architecture. Unfortunately, this complicates our builds, as the conda-packages are only working for `x86_64`/`AMD64`. Once we can get things working again for the `Raven-Hydro` conda package, that should be resolved.

There was also a recent update to `scikit-build-core` that needs to be addressed upstream as well. Working on that now.

PR #363, opened 2024-05-07
Pin pyogrio

Zeitsperre:


tlogan2000 (approved):
LGTM as long as test suite passes

Zeitsperre:
This is really only a patch release. There's a possibility that the fix for geopandas will be coming in the following days. We can address that alongside the new Raven-hydro package at the same time.

PR #372, opened 2024-05-31
Fallback in test if raven_hydro is not installed

### LINKED ISSUE #371: Allow users to override installing the `raven-hydro` dependency
The way that the requirements work right now, we absolutely need `raven-hydro` to be installed (either from PyPI/`pip` or via `conda`). Perhaps there could be a low-effort and safe approach to override this requirement for some special cases?

huard:
Fixes #371 

TODO: review dependencies to make raven_hydro a "soft" dependency ?

### DIFF
--- a/tests/test_emulators.py
+++ b/tests/test_emulators.py
@@ -1,4 +1,5 @@
 import datetime as dt
+import warnings
 
 import numpy as np
 import pytest
--- a/tests/test_emulators.py
+++ b/tests/test_emulators.py
@@ -8,7 +8,10 @@ from packaging.version import Version
 try:
     from raven_hydro import __raven_version__
 except ImportError:
-    warnings.warn(UserWarning, "The `raven-hydro` library is not installed. Setting '__raven_version__' to '0'.")
+    warnings.warn(
+        UserWarning,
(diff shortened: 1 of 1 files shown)


Zeitsperre:
> TODO: review dependencies to make raven_hydro a "soft" dependency ?

There's no rush just yet. Let's finish updating `raven-hydro` first and see where it makes the most sense to add the override.

tests/test_emulators.py
```diff
@@ -4,7 +4,11 @@
 import pytest
 import xarray as xr
 from packaging.version import Version
-from raven_hydro import __raven_version__
+
+try:
+    from raven_hydro import __raven_version__
+except ImportError:
+    __raven_version__ = "0"
```
Zeitsperre, line 11: How would you feel if we were to add a warning here?
```suggestion
try:
    from raven_hydro import __raven_version__
except ImportError:
    warnings.warn(UserWarning, "The `raven-hydro` library is not installed. Setting '__raven_version__' to '0'.")

    __raven_version__ = "0"
```

Zeitsperre (approved):

PR #394, opened 2024-09-18
Remove requests library, fix pyogrio calls, synchronize deps

Zeitsperre:
FYI @tlvu

### Changes

- Removed the pin on `pyogrio` (set by `geopandas` now).
- Set base required `geopandas` to v1.0.
- Updated remote requests code to rely on `urllib{3}` instead of `requests`.
- Updated calls to remote GeoJSON file-like objects.

### DIFF
--- a/src/ravenpy/utilities/geoserver.py
+++ b/src/ravenpy/utilities/geoserver.py
@@ -16,14 +16,13 @@ We could have a function that returns the layer name, and then other functions e
 import inspect
 import json
 import os
-import urllib.request
 import warnings
 from collections.abc import Iterable, Sequence
 from pathlib import Path
 from typing import Optional, Union
-from urllib.parse import urljoin
+from urllib.parse import urlencode, urljoin
 
-from requests import Request
+import urllib3
 
 from . import gis_import_error_message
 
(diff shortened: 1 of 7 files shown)


huard (approved):

src/ravenpy/utilities/geoserver.py
```diff
@@ -52,9 +51,12 @@
 
 # Can be set at runtime with `$ env RAVENPY_GEOSERVER_URL=https://xx.yy.zz/geoserver/ ...`.
 # For legacy reasons, we also accept the `GEO_URL` environment variable.
+HOST_URL = os.getenv(
+    "RAVENPY_HOST_URL", os.getenv("HOST_URL", "https://pavics.ouranos.ca/")
+)
 GEOSERVER_URL = os.getenv(
     "RAVENPY_GEOSERVER_URL",
-    os.getenv("GEO_URL", "https://pavics.ouranos.ca/geoserver/"),
+    os.getenv("GEO_URL", f"{HOST_URL}/geoserver/"),
```
tlvu, line 59: What is the default value of `HOST_URL`?  It must be the production `https://pavics.ouranos.ca/` and not the current host that can be any test servers which lack data !   Only the production server will be guarantied to have data populated.

Test servers that actually have the data can override `GEO_URL` or `RAVENPY_HOST_URL` to set it to its local Geoserver, otherwise the default must be the production Geoserver at https://pavics.ouranos.ca/geoserver/.

tlvu: Furthermore, I think RavenPy could be used as a library installed on any end user machine so the default Geoserver must be the production one, not the local one on the end user's machine (I assume `HOST_URL` default is the localhost, if not sorry for the false assumption).

PR #464, opened 2025-02-25
Add OpenSSF Best Practices Badge, Address Warnings

Zeitsperre:
### Pull Request Checklist:
- [ ] This PR addresses an already opened issue (for bug fixes / features)
  - This PR fixes #xyz
- [ ] (If applicable) Documentation has been added / updated (for bug fixes / features).
- [ ] (If applicable) Tests have been added.
- [x] CHANGELOG.rst has been updated (with summary of main changes).
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

* Adds the OpenSSF Best Practices badge (compliance is at 100%!)
* Addresses several warnings from deprecated usages.

### Does this PR introduce a breaking change?

It should not.

### Other information:

https://www.bestpractices.dev/en/projects/10064


src/ravenpy/config/rvs.py
```diff
@@ -512,11 +514,11 @@ def zip(
         return zip
 
 
-def is_symbolic(params: dict) -> bool:
+def is_symbolic(params: Union[dict, Any]) -> bool:
     """Return True if parameters include a symbolic variable."""
     from pymbolic.primitives import Variable
 
-    if is_dataclass(params):
+    if is_dataclass(params) and not isinstance(params, type):
```
huard, line 521: Why is this change needed ?

tests/test_forecasting.py
```diff
@@ -75,7 +75,7 @@ def test_forecasting_GEPS(numeric_config, get_local_testdata):
     # Prepare a RAVEN model run using historical data, GR4JCN in this case.
     # This is a dummy run to get initial states. In a real forecast situation,
     # this run would end on the day before the forecast, but process is the same.
-    wup.start_date = dt.datetime(2000, 1, 1)
```
huard, line 78: Fine as long as we don't need datetime.date or another object from the module. I've been bitten in the past, hence now suggest using dt for the module and not the class.

huard (approved):

src/ravenpy/config/rvs.py
Zeitsperre: It's a membership check to ensure that we pass an instance of the class and not the `dataclass` class type. The justification is that the following operation iterating over `params` will work for a `dataclass` instance, but not the `dataclass` type.

tests/test_forecasting.py
Zeitsperre: I think I added this while fixing some tests. This one is worth reverting. Thanks

PR #503, opened 2025-05-22
Fix-502

### LINKED ISSUE #502: 'overwrite=False' does not work in Emulator.run()
### Setup Information

- ravenpy version: 18.2

### Description

_No response_

### Steps To Reproduce

In `ravenpy.ravenpy.run()`, we have: 
https://github.com/CSHS-CWRA/RavenPy/blob/4652e4130b843a9baaf7437994c8a7488ea5cea1/src/ravenpy/ravenpy.py#L294-L307

Meaning that if `overwrite=False`, we don't delete the `output` folder, but we don't prevent Raven from overwriting the files.

Solution:
```
    if outputdir.exists():
        if overwrite:
            shutil.rmtree(str(outputdir))
        else:
            raise FileExistsError(...
```

### Additional context

_No response_

### Contribution

- [x] I would be willing/able to open a Pull Request to address this bug.

RondeauG:
### Pull Request Checklist:
- [x] This PR addresses an already opened issue (for bug fixes / features)
  - This PR fixes #502 
- [x] (If applicable) Documentation has been added / updated (for bug fixes / features).
- [x] (If applicable) Tests have been added.
- [x] CHANGELOG.rst has been updated (with summary of main changes).
  - [x] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

* `run(overwrite=False)` will now raise an error if output files already exist.
* `overwrite=True` will now only delete the output files related to the `modelname`, not the whole output folder.


src/ravenpy/ravenpy.py
```diff
@@ -296,8 +296,15 @@ def run(
...
-        shutil.rmtree(str(outputdir))
+    results = [f for f in outputdir.rglob(f"{modelname}_*.*")]
+    if len(results) > 0:
+        if overwrite:
+            for f in results:
+                Path(f).unlink()
+        else:
+            raise FileExistsError(
+                f"Output files already exist in {outputdir}. Use `overwrite=True` to overwrite."
+            )
```
Zeitsperre, line 307:
```suggestion
        else:
            msg = f"Output files already exist in {outputdir}. Use `overwrite=True` to overwrite."
            raise FileExistsError(msg)
```

src/ravenpy/ravenpy.py
```diff
@@ -296,8 +296,15 @@ def run(
     if not outputdir.is_absolute():
         outputdir = (configdir / outputdir).absolute()
 
-    if overwrite and outputdir.exists():
-        shutil.rmtree(str(outputdir))
+    results = [f for f in outputdir.rglob(f"{modelname}_*.*")]
```
Zeitsperre, line 299: Files always start with the model name?

RondeauG: If `modelname` is provided, then yes. I've added an option for when it isn't (e.g. `Hydrographs.nc`).

huard (changes requested):
I think we need to account for the RunName explicitly here. Fix comming.

tests/conftest.py
```diff
@@ -25,7 +25,7 @@
 )
 
 RAVEN_TESTING_DATA_BRANCH = os.getenv("RAVEN_TESTING_DATA_BRANCH", "master")
-SKIP_TEST_DATA = os.getenv("RAVENPY_SKIP_TEST_DATA")
+SKIP_TEST_DATA = True  # os.getenv("RAVENPY_SKIP_TEST_DATA")
```
Zeitsperre, line 28:
```suggestion
SKIP_TEST_DATA = os.getenv("RAVENPY_SKIP_TEST_DATA")
```

huard (approved):

PR #533, opened 2025-08-28
infer dimension ordering within nc_specs using cf-xarray

huard:
### Pull Request Checklist:
- [ ] This PR addresses an already opened issue (for bug fixes / features)
  - This PR fixes #xyz
- [ ] (If applicable) Documentation has been added / updated (for bug fixes / features).
- [ ] (If applicable) Tests have been added.
- [ ] CHANGELOG.rst has been updated (with summary of main changes).
  - [ ] Link to issue (:issue:`number`) and pull request (:pull:`number`) has been added.

* ...


RondeauG (approved):
LGTM!

So if I get this correctly, this will only help the generation of the `rvt` file so that the dimensions are in the expected order, right?

From my tests, Raven is also quite sensitive about the order of the dimensions within the dataset itself (t, y, x). Do you think that it is within the scope of RavenPy to warn the user if their data isn't ordered correctly?

RondeauG:
Actually, while I was talking about this with @richardarsenault, I realized that it might be a partial issue. With data in the X,Y,T order, I had this issue with missing meteorological data that proves that Raven still understood the NetCDF file correctly:
image

However, this was with fake elevation data that covered the entire grid. When I moved on to real elevation data (and fixing the issue of missing grid cells), I had errors that make me think that `elevation` specifically is still hard-coded `Y,X`.

huard:
I see explicit logic around the dimension order of lat, lon and elev. 
https://github.com/CSHS-CWRA/RavenHydroFramework/blob/f9c77a2491ce271493ea9629859d45f84ed181cd/src/ForcingGrid.cpp#L1016

